Issue #17, opened 2021-02-28
Create `common-hero.svelte` section 😃


EthanThatOneKid:
Create a file, `src/components/sections/common-hero.svelte`. This file will store the component that exists at the top of all of the subpages, i.e.: about, paths, node-buds.

The component should take a few properties: `title` and `picture` where `title` is the name of the current subpage and the `picture` property will be the source of the image used as the subpage's main graphic. Additionally, the component should optionally take a `headline` and/or `text` property. The `headline` will be the following headline on the page and the `text` will be the paragraph that follows the headline. Refer to the [Figma design](https://www.figma.com/file/9cvuO69WgNGuCjf2JGDPfq/ACM-Website-Mockup---Mike-Ploythai) to see how the component should render.

EthanThatOneKid:
Instead of properties, maybe we should use [_named slots_](https://svelte.dev/docs#slot_name) in order to retain the custom styling of the text passed to this component.

EthanThatOneKid:
#37

PR #37, opened 2021-03-20
Created `common-hero.svelte`

EthanThatOneKid:
I created the common-hero component and applied it to the about page. Soon, let's apply it to the other subpages.

This pull request is in response to #17.

### DIFF
--- a/src/components/sections/common-hero.svelte
+++ b/src/components/sections/common-hero.svelte
@@ -5,7 +5,7 @@
 
 <section>
   <header>
-    <object type="image/png" data="{src}" title="{alt}"></object>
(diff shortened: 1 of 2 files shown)


src/components/sections/common-hero.svelte
```diff
@@ -0,0 +1,78 @@
+<script lang="ts">
+  export let src: string = "../assets/png/acm-csuf-badge.png";
+  export let alt: string = "ACM CSUF Logo";
```
diamondburned, line 3: Using `const` here would probably be a better idea than `let`.

diamondburned (approved):

Issue #27, opened 2021-03-11
Create `why-join-nodebuds.svelte` section 👶


EthanThatOneKid:
This component is pretty simple. There is an image on the left with an title and paragraph to the right with right-justified text.

EthanThatOneKid:
#41

PR #41, opened 2021-03-26
Update nodeBuds path

bjwarfield:
added why-join-nodeBuds component
added how-can-i-get-involved component
modified nodebuds path to contain these components

addresses #27


EthanThatOneKid (approved):
The design of these components look good to me. We will need to migrate the SCSS code to CSS eventually, but it can be merged now and updated at a later date.

PR #42, opened 2021-04-09
Added the `acm-paths.svelte` section

EthanThatOneKid:
Right now, it is rendered in mobile mode only. The [`acm-paths.svelte` component](https://github.com/EthanThatOneKid/acmcsuf.com/blob/main/src/components/sections/acm-paths.svelte) can easily be updated for the desktop layout at a later date.


samuelsandoval1 (approved):
LGTM

Issue #29, opened 2021-03-11
Create `officer-profile-list.svelte` 📃


EthanThatOneKid:
This component will take raw officer data, the `acm-select.svelte` component, and the `officer-profile.svelte` component to render a grid of the officers that becomes a column under mobile dimensions.

EthanThatOneKid:
Completed in #43 😎

Issue #24, opened 2021-03-09
Create `acm-select.svelte` ▶


EthanThatOneKid:
Create a file, `/src/components/ui/acm-select.svelte` that contains the drop-down select component on the *about* page.

Here is the markup that has already been created:

```html
<select name="school-year" bind:value="{currentSchoolYear}">
  {#each Object.keys(officers) as schoolYear (schoolYear)}
    <option value="{schoolYear}">{schoolYear}</option>
  {/each}
</select>
```

This component will need one properties and one event type. The property will be `export let datalist: string[];` (which will take the place of `students` in the above code). The event will be called `"change"`. Here is some [info on making an event dispatcher in Svelte](https://svelte.dev/docs#createEventDispatcher).

Also, here is the current associated styling:

```scss
select {
  background-color: $acmDark;
  padding: 1em;
  color: $acmLight;
  font-family: "Poppins", sans-serif;
  font-weight: bold;
  border-radius: 25px;

  option {
    background-color: $acmDark;
    color: $acmLight;
    border-radius: 25px;
  }
}
```

This styling was written with SCSS, but this component will be rewritten in CSS. To convert to CSS, replace `$acmDark` with `var(--acm-dark)` and replace `$acmLight` with `var(--acm-light)`.

EthanThatOneKid:
Completed in #43 😎

PR #43, opened 2021-04-09
Created the `officer-profile-list.svelte` component and more.

EthanThatOneKid:
I created the necessary `officer-profile-list.svelte` as well as the `acm-select.svelte` UI component to go with it.


EthanThatOneKid:
This pull request is intended to close both issue #29 and #24 since the two efforts ended up relying on each other.

Issue #13, opened 2021-02-27
Replace fluidSize mixin instances with mobile-first CSS


EthanThatOneKid:
A SCSS mixin is used in many Svelte files across the project. I used it to help with implementing the design, but they really should all be converted into mobile-first CSS solutions for better maintainability and structure.

Example:

```scss
@include fluidSize(18, 48, $propName: "font-size");
```

Here is an example of fluidSize being used. 18 is the value in pixels for the mobile design and 48 is the value in pixels for the desktop design. `"font-size"` is the CSS property that is being targeted. A mobile-first CSS alternative is more long-winded, but will be more maintainable over time.

```css
p {
  font-size: 18px;
}

@media screen and (min-width: $breakpoint) {
  p {
    font-size: 48px;
   }
}
```

> 💡 Use your best judgement when messing around with the CSS.

diamondburned:
Is falling back to `@media` for `font-size` a really good idea? Perhaps `em` could be used as a relative font size instead, combined with the [viewport meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag), which wouldn't need edge cases with `@media`.

EthanThatOneKid:
I agree that the `font-size` property should use `em` units 👍.

The above code is just an example for converting from the `fluidSize` mixin in general. In the codebase, there are instances of `fluidSize` where this conversion would make sense.
For example, when the property is `width` instead of `font-size`.

diamondburned:
It seems like a CSS `var` can't be used on a `@media` like so:

```css
  @media screen and (min-width: var(--desktop-breakpoint)) { /* <-- does not work */
	.screen-wyd div {
	  padding-left: var(--max-padding);
	  padding-right: var(--max-padding);
	  padding-bottom: var(--max-p-font-size);
	}

	.screen-wyd p,
	.screen-wyd h2 {
	  font-size: var(--max-h-font-size-med);
	  padding-bottom: var(--max-p-font-size);
	};
  };
```

I suppose we still have to keep that specific part SCSS, though I'm not sure if I should still convert all other variables to CSS.

diamondburned:
In the future, I think it'll be better if we move all the common attributes like `font-size: var(--min-p-font-size)` to `global.css`.

PR #52, opened 2021-04-10
Fix nodeBuds paragraph alignment, turnback class

### LINKED ISSUE #49: Fix some CSS bugs in `why-join-nodebuds.svelte` section
Specifically, the text needs to be tweaked so that it is aligned properly in both mobile and desktop dimensions.

Here is the [file in question](https://github.com/EthanThatOneKid/acmcsuf.com/blob/main/src/components/sections/why-join-nodebuds.svelte). Happy bug-fixing 😊!

diamondburned:
Fix nodeBuds paragraph alignment, turnback class
    
    This commit fixes the second paragraph's alignment in mobile view to
    render properly. It also fixes a minor capitalization error ("nodebuds"
    to "nodeBuds").
    
    This commit also adds the groundwork for potential future refactoring:
    the CSS variables --fluid-font-size, --p-font-size and --h-font-size-med
    were introduced to slowly replace the previous fluidSize mixin.
    
    The --fluid-font-size uses the CSS clamp() math function to achieve
    smoothly scalable font sizes across all viewports unlike setting manual
    font-sizes over @media. This behavior is almost similar to the mixin.
    Its intended use case is to be set at the parent container (or the
    body). Note that this affects the margins and paddings if they use em as
    well.
    
    The two new font size variables, --p-font-size and --h-font-size-med,
    are in em units to properly utilize the above --fluid-font-size set in
    the parent container. The em units will scale accordingly. This is
    different from using px, which is scaled absolutely (as opposed to
    relatively to the parent's font-size).

---

This PR closes #49.

Certain note-worthy things:

1. The old `<div class="turnback">` that wraps everything is removed.
2. `<div class="turnback">` is only used to wrap the text paragraph now.
3. `.turnback` CSS code should be OK to be copy-pasted into `global.css` once other components remove its own (probably).

### DIFF
--- a/src/components/sections/why-join-nodebuds.svelte
+++ b/src/components/sections/why-join-nodebuds.svelte
@@ -1,82 +1,89 @@
-<div class="turnback">
-  <div class="why-join-nodebuds">
-    <!-- <img src="../assets/png/why-join-node-buds-illustration.png" alt="" /> -->
-
-    <object
-      type="image/png"
-      data="../assets/png/why-join-node-buds-illustration.png"
-      title="Why Join nodeBuds"></object>
-    <div>
-      <h2>why join node<span class="brand-red">buds</span>?</h2>
-      <p>
-        <span class="brand-em">node<span class="brand-red">Buds</span></span>
-        was designed to provide underclassmen and transfer students assistance as
-        they start their fresh journey through
-        <span class="brand-blue brand-em">CSUF</span>, and through the tech
-        industry. We provide the knowledge and resources that’ll help students
-        start on the right foot.
-      </p>
-    </div>
+<div id="why-join-nodebuds">
+  <img
+    src="../assets/png/why-join-node-buds-illustration.png"
+    alt="Why Join nodeBuds"
+    title="Why Join nodeBuds"
+  />
+  <div class="turnback">
+    <h2>why join node<span class="brand-red">Buds</span>?</h2>
+    <p>
+      <span class="brand-em">node<span class="brand-red">Buds</span></span>
+      was designed to provide underclassmen and transfer students assistance as they
+      start their fresh journey through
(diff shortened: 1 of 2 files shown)


diamondburned:
Relevant issue: https://github.com/EthanThatOneKid/acmcsuf.com/issues/13, also to do with the `fluidSize` mixin refactor.

src/components/sections/why-join-nodebuds.svelte
```diff
@@ -1,82 +1,89 @@
...
+    #why-join-nodebuds > img {
+      width: max(55vw, 550px);
+      max-width: 100%;
+    }
+    #why-join-nodebuds > .turnback > p {
+      max-width: max(30vw, 300px);
+    }
+
+    .turnback,
+    .turnback:nth-child(even) {
```
EthanThatOneKid, line 82: Here, the `.turnback:nth-child(even)` selector is possibly unnecessary since the `.turnback` selector should work for all `.turnback:nth-child(even)` elements.

EthanThatOneKid (commented):
Overall these changes look great! I am looking forward to dropping in these utility classes from the `global.css` file.

Additionally, thank you for the very detailed commit message. It was a good read.

src/components/sections/why-join-nodebuds.svelte
diamondburned: I actually tested this, and without the `:nth-child(even)`, the outer one would take priority over this one.

EthanThatOneKid: Ah that makes sense. Thank you for looking into that. 😀

That would align with CSS's concept of [_specificity_](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity).

Issue #53, opened 2021-04-10
Update `nodebuds-testimonial.svelte` section's CSS


EthanThatOneKid:
The current implementation only accounts for mobile dimensions. Preferably, some CSS should be added to represent the desktop layout. A media query such as `@media screen and (min-width: 756px) { /* ... */ }` will be needed to account for the desktop dimensions.

In this case, the image should be positioned to the right of the text under desktop dimensions.

PR #58, opened 2021-04-12
Update Nodebuds sections

bjwarfield:
Add desktop css layout for nodebulds-testimonial
Adjusted CSS between nodebuds-testimonial and why-join-nodebuds to have similar breakpoints and spacing.
Adjusted top and bottom padding for how-can-I-get-involved

addresses #53


EthanThatOneKid (approved):
These changes look great to me. Nice use of `flex-direction: column-reverse;` as well as the CSS variables for the font sizes.

Issue #56, opened 2021-04-10
Update all "section" components to account for the newly added `Spacing` component 🌌


EthanThatOneKid:
Most of the components within `src/components/` are responsible for their own spacing between each other. The responsibility has fallen onto the `Spacing` component which is used amongst the other section components and can be configured to have custom spacing where needed. The goal for this task is to identify where spacing has been used (i.e. `margin-top`, `padding-bottom`, etc.) and remove it. The `Spacing` component has already been added where it needs to be added, so all there is to do is remove the old, unnecessary spacing between sections.

- Components in question: <https://github.com/EthanThatOneKid/acmcsuf.com/tree/main/src/components/sections>
- The `Spacing` component: <https://github.com/EthanThatOneKid/acmcsuf.com/blob/main/src/components/sections/spacing.svelte>

Issue #45, opened 2021-04-09
Improve `acm-paths.svelte` 🖥


EthanThatOneKid:
The [`acm-paths.svelte`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/main/src/components/sections/acm-paths.svelte) component was developed with mobile dimensions in mind. By adding media querie(s) to render the desktop-size design, this component can be improved.

> 💡 Note: This task will involve changing the column of path items into a row with `flex-direction: row;`.

EthanThatOneKid:
Closed after pull request #60.

PR #60, opened 2021-04-15
Meeting with Designer

EthanThatOneKid:
The changes in this commit are the result of the notes of a meeting with the site's designer (@mikeploythai). The main goal of this meeting was to accurately implement the design of the index page.

## Structural Changes

Component sections of the index page (that are not used anywhere besides the index page) have been moved to a new directory, `/src/components/index`. This includes `acm-paths.svelte`, `hero.svelte`, and `why-join.svelte`.

For the NodeBuds page, some unique section components were also moved into a new directory, `/src/components/nodebuds`. This includes `nodebuds-testimonial.svelte` and `why-join-nodebuds.svelte`. Take note that  I also changed the URL of the NodeBuds endpoint from `node-buds` to `nodebuds` since it is written as one word.

Additionally, the `src/components/ui` directory has been renamed to `src/components/utils`.

## Stylistic Changes

New variables have been added to the `global.css` file! The following variables have to do with the `font-size` style property and are intended to be used everywhere throughout the site.

```css
:root {
  --heading-font-size: 48px;
  --subheading-font-size: 24px;
  --body-font-size: 18px;
  --margin-text-font-size: 12px;
}

@media screen and (min-width: 768px) {
  :root {
    --title-font-size: 96px;
    --heading-font-size: 64px;
    --subheading-font-size: 36px;
    --body-font-size: 24px;
    --margin-text-font-size: 16px;
  }
}
```

The values of the variables outside of the media query are the values used in mobile mode. The values set inside of the media query are the values used in tablet/desktop mode. All instances in which the `font-size` style property is adjusted on the index page have been updated to use the newly added CSS variables above.

I also added another CSS variable, `--navbar-height`, to `global.css` which is used by the `navbar.svelte` component to set the height of the navbar.

In a number of files that are used on the index page, any unnecessary padding has been eliminated in favor of the `Spacing` component. This change contributes to issue #56, but does not complete it.

Also, all of the instances on the index page where some sort of max-width is used (as a constraint to help with the page's layout) were replaced with margins instead.

In order to achieve the goal of implementing the design of the index page, issue #45 had to be tackled as part of this commit. The `acm-paths.svelte` component now becomes a row when in tablet/desktop mode.


src/components/index/why-join.svelte
```diff
@@ -9,60 +9,60 @@
   <div>
     <h2>why should students join acm?</h2>
     <p>
-      <span class="bold">acm<span class="blue">CSUF</span></span>
+      <span class="brand-em">acm<span class="brand-em">CSUF</span></span>
```
diamondburned, line 12: Why is there 2 nested spans with the same class? I'm assuming the one inside should have a different class?

static/global.css
```diff
@@ -1,5 +1,9 @@
 @import url("https://fonts.googleapis.com/css2?family=Poppins:wght@400;600;700");
 
+html, body {
+  overflow-x: hidden;
```
diamondburned, line 4: This isn't important, but I think that if we need this, then we have other components to fix instead of adding `overflow: hidden`.

src/components/index/why-join.svelte
ghost: The current code assumes that the entire naming scheme is entirely bold.  This is incorrect.

With the current design guidelines, the only thing that should have font-weight: bold, are the brandings.  This includes, and is exclusive to, the words, "CSUF", "Dev", "Create", "Algo", and "Buds".  Every other character in the title should have font-weight: 600.  This property also applies to headings and subheadings.  This is a very easy fix that I can send a pull request in for later.

EthanThatOneKid: Thank you for noticing my oversight, @diamondburned. I created an issue (<https://github.com/EthanThatOneKid/acmcsuf.com/issues/62>) on the project board for you to submit your fixes onto, @mikeploythai.

static/global.css
EthanThatOneKid: I agree that we have other components to fix on the index page. As of right now, the layout is pretty solid for once using the `overflow-x: hidden;` style which is why I left it here temporarily. I have created issue #63 to look into what is causing the horizontal scroll on smaller screen sizes.

PR #84, opened 2021-04-24
Added Google Form option for the `/connect` page

EthanThatOneKid:
I was not able to get access to acmCSUF's official MailChimp account, so we cannot use the HTML form that is shown in the design currently. Once we get access to the MailChimp account, we can easily go back to the HTML form that @bjwarfield created!

TL;DR: As a temporary alternative, we are embedding a Google Form.

### DIFF
--- a/src/routes/connect.svelte
+++ b/src/routes/connect.svelte
@@ -4,6 +4,8 @@
   import SocialMediaLinks from "@/components/sections/social-media-links.svelte";
   import GetInTouchForm from "@/components/sections/get-in-touch-form.svelte";
   import CallToActionSection from "@/components/sections/call-to-action-section.svelte";
+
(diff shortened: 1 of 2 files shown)


src/routes/connect.svelte
```diff
@@ -5,7 +5,8 @@
   import GetInTouchForm from "@/components/sections/get-in-touch-form.svelte";
   import CallToActionSection from "@/components/sections/call-to-action-section.svelte";
 
-  const googleFormUrl = `https://docs.google.com/forms/d/e/1FAIpQLSfJanOAaL2mdjpf193tFeCClBzpW_COEO_crAE8hqsJCB_Rwg/viewform?embedded=true`;
+  const googleFormId =
+    "1FAIpQLSfJanOAaL2mdjpf193tFeCClBzpW_COEO_crAE8hqsJCB_Rwg";
```
EthanThatOneKid, line 9: The form ID itself can also be stored in a constants file elsewhere in the project and imported by the `get-in-touch-form.svelte` component. But, by passing the ID as a property does make the component reusable.

samuelsandoval1 (approved):
LGTM

diamondburned (approved):

PR #91, opened 2021-04-24
Event Page Progress 🏔

### LINKED ISSUE #75: Create the groundwork for the `/events` page 📆
This page will include adding a `preload` function to the Svelte file's script with the module typing. This includes starting on #72 and #74.

Read #20 for more information on fetching Google Calendar data.

### LINKED ISSUE #90: Refactor the `event-carousel.svelte` 📆
We need to create some JavaScript-powered interactions for the carousel.

Read: #91

EthanThatOneKid:
We created the layout for the `/events` page and rendered the ICAL data from the Google Calendar.

We also fixed some styles that conflicted with the event card that originated in the `call-to-action.svelte` component, so we added those fixes where needed.

- [x] We need a button on either side to increment the scroll-offset slightly in its respective direction.
- [x] We need to make it more mobile-friendly, so let's add some media queries.
- [x] Create a drag gesture/interaction to increment the scroll-offset smoothly. (<https://github.com/EthanThatOneKid/acmcsuf.com/pull/91/commits/1bf4961b34369e20e8c27f782bf027a283bcb807>)
- [x] Automatically scroll to event with the URL's _fragment_.
- [x] Hide buttons for mobile dimensions.
- [x] Hide the scrollbar.


EthanThatOneKid:
I was looking into ways to accomplish the fading effect for the horizontal edges of the event carousel. One possible solution I've found is [CSS linear gradients](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient()). With linear gradients, we could set the background image of an overlayed element to two linear gradients.

The definition of the linear gradients would look something like the code snippet below.

```css
background: linear-gradient(to right, var(--transparent) 75%, var(--acm-light)),
  linear-gradient(to left, var(--transparent) 75%, var(--acm-light));
```

src/components/events/event-carousel.svelte
```diff
@@ -1,13 +1,52 @@
 <script lang="ts">
+  import { onMount } from "svelte";
   import EventItem from "./event-item.svelte";
   import type { AcmEvent } from "../../lib/parse-ical-data";
 
+  enum GrabbingState {
```
diamondburned: Do we need this? Why can't we do `grabbing bool`?

EthanThatOneKid: I am using this enum because it is being used as an attribute value. I assume there is a simpler way to do this involving a boolean. In the end, the CSS just needs to target something to be able to give it the appropriate style which could be a class or an attribute.

We could possibly instead use an `isGrabbing` variable and use it to assign a class onto the div using a Svelte class directive like this: `<div class:grabbing={isGrabbing} ... />`.

EthanThatOneKid: Addressed in commit <https://github.com/EthanThatOneKid/acmcsuf.com/pull/91/commits/b2d9ffd220775905eb63197f599820092cbffe6f>.

EthanThatOneKid:
Continued efforts on #99 due to an early merge of this pull request.

PR #105, opened 2021-05-01
Serving iCal data for client to fetch on page load

EthanThatOneKid:
This will allow to render the event data every time the user enters the page.

### DIFF
--- a/src/lib/parse-ical-data.ts
+++ b/src/lib/parse-ical-data.ts
@@ -44,7 +44,7 @@ export const parseIcalData = (icalData: string): AcmEvent[] => {
       const slug = slugifyEvent(summary, month, day);
       return { month, day, time, location, summary, date, endDate, isHappening, slug };
     })
-    // .filter(({ endDate }) => endDate.valueOf() > now) // Comment out this filter statement to show a longer list of events for testing purposes.
+    .filter(({ endDate }) => endDate.valueOf() > now) // Comment out this filter statement to show a longer list of events for testing purposes.
     .sort(({ date: date1 }, { date: date2 }) => date1.valueOf() > date2.valueOf() ? 1 : -1)
   return jcalData;
 };
--- a/src/lib/parse-ical-data.ts
+++ b/src/lib/parse-ical-data.ts
@@ -16,7 +16,14 @@ export interface AcmEvent {
   location: string;
   summary: string;
   isHappening: boolean;
+  slug: string;
 }
+
+const slugifyEvent = (summary: string, month: string, day: number): string => {
+  const cleanedSummary = summary.replace(/[^\w\s\_]/g, "").replace(/(\s|\-|\_)+/g, "-");
+  const slug = [cleanedSummary, month, day].join("-").toLowerCase();
+  return slug;
+};
 
 export const parseIcalData = (icalData: string): AcmEvent[] => {
   const now = Date.now();
@@ -34,7 +41,8 @@ export const parseIcalData = (icalData: string): AcmEvent[] => {
       const day = date.getDate();
       const time = date.toLocaleTimeString(acmLocale, { hour: "numeric", minute: "numeric" });
       const isHappening = now >= date.valueOf() && now < endDate.valueOf();
-      return { month, day, time, location, summary, date, endDate, isHappening };
+      const slug = slugifyEvent(summary, month, day);
+      return { month, day, time, location, summary, date, endDate, isHappening, slug };
     })
     .filter(({ endDate }) => endDate.valueOf() > now) // Comment out this filter statement to show a longer list of events for testing purposes.
     .sort(({ date: date1 }, { date: date2 }) => date1.valueOf() > date2.valueOf() ? 1 : -1)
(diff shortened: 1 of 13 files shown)


EthanThatOneKid:
Let's use this code from [ical2json](https://github.com/adrianlee44/ical2json) to replace `ical.js`. We just need to make sure that the function/file containing the pasted data attributes the author.

<https://github.com/adrianlee44/ical2json/blob/main/src/ical2json.ts>

src/lib/parse-ical-data.ts
```diff
@@ -18,25 +71,41 @@ export interface AcmEvent {
   isHappening: boolean;
 }
 
+const convertIcalDateTime = (datetime: string): Date => {
+  const characters = datetime.split("");
+  const fullYear = characters.splice(0, 4).join("");
```
diamondburned: `characters.splice(a, b).join("")` could be `datetime.slice(a, b)` instead without splitting `datetime`.

src/routes/events.json.ts
```diff
@@ -0,0 +1,22 @@
...
+    "https://calendar.google.com/calendar/ical/738lnit63cr2lhp7jtduvj0c9g%40group.calendar.google.com/public/basic.ics";
+
+const expirationTimeout = 1e3 * 60 * 30; // Fetch updates every 30 minutes.
+let eventExpirationTimestamp = 0;
+let events: AcmEvent[] = [];
+
+export async function get(req: Request, res: Response) {
+    const now = Date.now();
+    if (now > eventExpirationTimestamp) {
+        const data = await fetch(ICAL_TARGET_URL).then(response => response.text());
```
diamondburned, line 16: Is this still done by the client? Does it need `no-cors` then?

src/routes/events.svelte
```diff
@@ -6,35 +6,20 @@
...
-    // attached to the URL, the scroll needs to be
-    // offset slightly upwards.
-    if (location.hash.length > 0) {
-      scrollBy({ top: -200 });
-    }
-
     // Lazily load the ICAL data for the event carousel.
-    fetch(ICAL_TARGET_URL, { mode: "no-cors" })
-      .then((response) => response.text())
+    fetch("../events.json")
```
diamondburned, line 15: Where is this pointed to?

EthanThatOneKid: I created a public API endpoint (to act as a proxy server) to see if that _could_ solve the problem. If you go to `acmcsuf.com/events.json`, it will load the array of events.

diamondburned: I'm getting a 404.

Would this require a server then?

src/lib/parse-ical-data.ts
EthanThatOneKid: That would possibly be more readable. It would indeed work since the numbers in the iCal datetime format are always in the same spots.

src/routes/events.json.ts
EthanThatOneKid: This is being ran on the server, so we can omit the `no-cors` mode.

src/lib/parse-ical-data.ts
EthanThatOneKid: Fixed: <https://github.com/EthanThatOneKid/acmcsuf.com/pull/105/files#diff-c9b7c8232c884e8d6c62a724e8cf309ba0d316ca3bf4cad5a1515820a0911edeR76-R81>

src/routes/events.svelte
EthanThatOneKid: Here is what it looks like based on the preview: <https://acm-csuf-site-cksb0sp2f-ethanthatonekid.vercel.app/events.json>

diamondburned:
Vercel being able to run server code makes me think we could probably not use `fetch` at all on the client and just have Vercel update the static page directly. This is nitpicking, of course, and it would make caching of the `/events` route less friendly as well.

diamondburned (approved):

EthanThatOneKid:
> Vercel being able to run server code makes me think we could probably not use `fetch` at all on the client and just have Vercel update the static page directly. This is nitpicking, of course, and it would make caching of the `/events` route less friendly as well.

By rendering the event page statically, I feel like there will be moments where the page is outdated. This outdated render of the page may even get cached (as you mentioned) and would be difficult for an average user to get around.

One advantage to rendering the page statically would be minimizing how many network requests are made between the server and the client. With the current solution, as long as the user keeps refreshing the page, they will continue making requests to the Vercel server which can be bad. I integrated a caching feature so that the Google Calendar data is only fetched once every 30 minutes ([number can be easily changed](https://github.com/EthanThatOneKid/acmcsuf.com/blob/59530e5021ef6edcf247da84985684cdaaf908bb/src/routes/events.json.ts#L9)). Another problem is that since the data is rendered after the client has fetched it, we need to run extra JavaScript to enable the highlighting feature we created.

I agree that we should look into rendering the page statically after weighing the pros and cons. The next step is figuring out how this would actually be implemented. We could possibly run a GitHub action every so often to make Vercel automatically rebuild the site every few minutes and that could solve the problem. Further discussion/planning is needed before we can get started with implementing this though.

PR #111, opened 2021-05-04
Updated `acm-button.svelte` component styling

EthanThatOneKid:
Implemented @mikeploythai's button design. Now, a smoother transition happens when the user hovers over any button element on the site.

Default style:

image

Style on hover:

image

### DIFF
--- a/src/components/utils/acm-button.svelte
+++ b/src/components/utils/acm-button.svelte
@@ -12,20 +12,25 @@
     cursor: pointer;
     border-radius: 15px;
     user-select: none;
-    background-color: var(--acm-dark);
+    background-color: none;
     padding: 8px 16px 8px 16px;
     margin-top: 24px;
     font-weight: bold;
-    color: var(--acm-light);
+    color: var(--acm-dark);
     text-decoration: none;
     border: 2px solid var(--acm-dark);
     font-size: var(--body-font-size);
     text-transform: lowercase;
+    box-shadow: 0 0 0 rgba(23, 26, 28, 0.1);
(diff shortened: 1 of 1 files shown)


src/components/utils/acm-button.svelte
```diff
@@ -12,20 +12,25 @@
...
   }
 
   .call-to-action:hover {
     transition: color 0.25s ease-in-out, background-color 0.25s ease-in-out;
-    color: var(--acm-blue);
-    background-color: var(--acm-light);
+    color: var(--acm-light);
+    background-color: var(--acm-dark);
+    -webkit-filter: drop-shadow(0 2px 10px rgba(23, 26, 28, 0.5));
+    filter: drop-shadow(0 2px 10px rgba(23, 26, 28, 0.5));
```
diamondburned: Does `box-shadow` not work? It should work with `border-radius`.

diamondburned (changes requested):

src/components/utils/acm-button.svelte
EthanThatOneKid: `box-shadow` probably works great. I was using the same properties as @mikeploythai's implementation, but `box-shadow` makes more sense here after further consideration. Resolved by commit <https://github.com/EthanThatOneKid/acmcsuf.com/pull/111/commits/e78d5e5f3747ea09896cf4b811a86335598bf6c5>.

diamondburned (approved):

bjwarfield (approved):

PR #113, opened 2021-05-06
Added "wheel" event to the carousel ♿

EthanThatOneKid:
The carousel now has a new event attached that waits for the user to use their scroll wheel to scroll over the event carousel. When the user scrolls down while hovering over the carousel, the carousel will move to the right.

### DIFF
--- a/src/components/events/event-carousel.svelte
+++ b/src/components/events/event-carousel.svelte
@@ -5,7 +5,7 @@
 
   export let events: AcmEvent[] = [];
 
-  const scrollIncrementDistance = 150;
+  const scrollIncrementDistance = 250;
   let carouselRef: HTMLDivElement;
   let carouselButtonLeft: HTMLDivElement;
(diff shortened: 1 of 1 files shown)


diamondburned (approved):

src/components/events/event-carousel.svelte
```diff
@@ -29,43 +28,38 @@
...
       {#each events as eventInfo}
         <EventItem info="{eventInfo}" />
       {/each}
       <div class="event-item-buffer"></div>
     </div>
-    <div bind:this="{carouselButtonRight}" class="carousel-button right">
+    <div
+      bind:this="{carouselButtonRight}"
+      class="carousel-button right"
+      on:click="{scrollLeft}"
```
diamondburned: This should probably be `scrollRight`

diamondburned (changes requested):

diamondburned (approved):

PR #116, opened 2021-05-11
Change CRLF to LF

### LINKED ISSUE #114: Replace all line breaks to LF 😵
This codebase was originally written on Windows and does not account for all of the ways that line breaks are read on all systems. To be safe, all line breaks should be encoded with LF.

See: <https://superuser.com/a/293951>
Read: <https://git-scm.com/docs/gitattributes#_eol>

diamondburned:
Command used: `dos2unix .vscode/** src/** static/** *`

This pull request should fix #114. It does not make a `.gitattributes` file, since editors should be smart enough to detect this.


EthanThatOneKid (approved):
Everything looks good! Thanks for the help, @diamondburned and `dos2unix` 😁

PR #138, opened 2021-09-24
Finalized v2 designs

ghost:
- every board member has an updated badge
- removed nodebuds testimonial for now until i receive more from aaron
- fixed font weights and sizes
- updated favicon
- event carousel should be showing properly on ios now
- updated discord logo


src/components/about/officer-profile-list.svelte
```diff
@@ -13,7 +13,7 @@
   };
 
   const formattedTerms = TERMS.map(formatTerm);
-  let currentTermIndex = TERMS.length - 1;
+  let currentTermIndex = TERMS.length - 2;
```
EthanThatOneKid: Set this to 0.

src/components/connect/social-media-links.svelte
```diff
@@ -3,26 +3,25 @@
   import DiscordIcon from "@/components/icons/discord.svelte";
   import LinkedInIcon from "@/components/icons/linkedin.svelte";
   import YouTubeIcon from "@/components/icons/youtube.svelte";
-  import { links } from "../../lib/links";
```
EthanThatOneKid, line 6: Good simplification here.

src/components/utils/admonition.svelte
```diff
@@ -4,16 +4,16 @@
   export let expiration: number;
 
   const isExpired = new Date().valueOf() > expiration;
-  const icon = `../assets/badges/${path}.svg`;
+  const icon = `../assets/badges/with-shadow/${path}.svg`;
```
EthanThatOneKid: Here, I'd keep the `with-shadow` out of this path. The component would be used something like `<Admonition picture="with-shadow/nodebuds.svg" />`.

src/routes/_layout.svelte
```diff
@@ -3,17 +3,16 @@
   import Footer from "@/components/sections/footer.svelte";
 
   export let segment: string;
-  const pageTitle = segment === undefined ? "ACM at CSUF" : segment;
 </script>
 
 <Navbar segment="{segment}" />
-<main><slot /></main>
+<main style="min-height: 100vh;"><slot /></main>
```
EthanThatOneKid: These styles would look good between some of these: `<style></style>`.

EthanThatOneKid (changes requested):
Just a few nitpicks, but great work! Thank you for combing through each file, fine-tuning everything.

src/components/about/officer-profile-list.svelte
ghost: setting it to 0 doesn't make the list appear until i select spring 2021, then fall 2021

image

src/components/connect/social-media-links.svelte
ghost: Wanted to make use of our custom URLs as a subtle touch to make the site more polished, but i'll reimplement them if you really want it

src/components/utils/admonition.svelte
ghost: Isn't that already what the path line is for?  We dont need to specify like that when we can do it automatically by presetting `with-shadow` in the path

src/routes/_layout.svelte
ghost: ok

src/components/connect/social-media-links.svelte
EthanThatOneKid: What you did works great to me 😁

src/components/utils/admonition.svelte
EthanThatOneKid: If we want the behavior to always use path badges, then the method you've proposed works great. However, we may want to use badges that do not use shadows. Or to use any other static image on the site.

The plan is to use this feature to automate announcements in which case we'd want to be able to customize which icon is used for the announcement.

src/components/utils/admonition.svelte
```diff
@@ -1,19 +1,19 @@
 <script lang="ts">
-  export let path: string = "general";
   export let text: string = "";
   export let expiration: number;
+  export let picture: string = "";
 
   const isExpired = new Date().valueOf() > expiration;
-  const icon = `../assets/badges/${path}.svg`;
+  const icon = picture;
```
EthanThatOneKid: Replace this with

```
const icon = `../assets/badges/${path}.svg`;
```

for now. This will work for both `path = "nodebuds"` _and_ `path = "with-shadows/nodebuds"`.

src/routes/about.svelte
```diff
@@ -39,21 +39,23 @@
 <!-- TODO: Get the content below from the `/events.json` endpoint and just
 display the most upcoming event with a reminder/notice flair. -->
 <Admonition
-  path="nodebuds"
+  picture="assets/badges/with-shadow/nodebuds.svg"
```
EthanThatOneKid: For now, all you need is "with-shadow/nodebuds".

EthanThatOneKid (changes requested):

EthanThatOneKid (approved):
Great work! LGTM 😎

PR #146, opened 2021-10-12
Fixed bug on NodeBuds officer section (wrong officers showing)

### LINKED ISSUE #141: Profile component bug for nodeBuds
In [nodeBuds page](https://acmcsuf.com/nodebuds), profile component will pass any past nodebuds officer to current year so long as they

1. previously had a nodeBuds role
2. is still in the board

EthanThatOneKid:
Fixed the NodeBuds officer bug, #141.

- Created a new Svelte store to keep the same `termIndex` on both `/about` and `/nodebuds`.
- Simplified `<OfficerProfileList />` by adding a `filter` property to it. This way the filter function is handled outside the component, allowing the component to re-render when the `termIndex` changes on another page.
- Implemented the necessary changes on both `/about` and `/nodebuds`.

Resolving #141.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,4334 @@
 {
   "name": "acmcsuf.com",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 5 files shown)


ghost (approved):
LGTM!

PR #153, opened 2021-10-19
Modified navbar.svelte such that menu will close when link clicked

### LINKED ISSUE #124: Menu does not close after page navigation 🧱
The mobile menu component currently does not close when a menu item is clicked. This requires the user to click the ❌ on the top right of the screen in order to close the menu.

The expected behavior, however, is to close the menu automatically when a menu item is clicked.

anhduy1202:
Hi, hope this is the correct way to make a PR, let me know if I need to improve anything in my code, thank you.

### DIFF
--- a/src/lib/components/sections/navbar.svelte
+++ b/src/lib/components/sections/navbar.svelte
@@ -1,34 +1,49 @@
 <script lang="ts">
 	export let segment: string;
+	let menu;
+	let checkbox;
+
 	const menuItems = [
 		{ title: 'about', slug: 'about' },
 		{ title: 'events', slug: 'events' },
 		{ title: 'paths', slug: 'paths' },
 		{
 			title: '<span class="headers">node</span>Buds',
-			slug: 'nodebuds'
+			slug: 'nodebuds',
 		},
-		{ title: 'connect', slug: 'connect' }
+		{ title: 'connect', slug: 'connect' },
 	];
+
+	const handleClose = () => {
+		if (checkbox.checked === true) {
+			menu.style.visibility = 'hidden';
+			checkbox.checked = false;
+		}
+	};
+	const handleOpen = () => {
+		if (checkbox.checked === false) {
+			menu.style.visibility = 'visible';
+		}
+	};
 </script>
 
 <nav>
-	<input type="checkbox" id="navToggle" />
+	<input type="checkbox" id="navToggle" bind:this={checkbox} />
 	<div class="content">
 		<a href="/" class="full-logomark">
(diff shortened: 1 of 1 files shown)


src/lib/components/sections/navbar.svelte
```diff
@@ -1,42 +1,57 @@
 <script lang="ts">
 	export let segment: string;
+	let menu;
+	let checkbox;
```
EthanThatOneKid: This is just a nitpick, but to be as explicit with our intentions with a variable as possible in TypeScript, we can define variables with their expected HTML element. Like so:

```ts
let menu: HTMLUListElement;
let checkbox: HTMLInputElement;
```

src/lib/components/sections/navbar.svelte
```diff
@@ -1,42 +1,57 @@
...
-			slug: 'nodebuds'
+			slug: 'nodebuds',
 		},
-		{ title: 'connect', slug: 'connect' }
+		{ title: 'connect', slug: 'connect' },
 	];
+
+	const handleClose = () => {
+		if (checkbox.checked === true) {
+			menu.style.visibility = 'hidden';
```
EthanThatOneKid: The neat thing about the technique used with the menu is that it does not actually need any JavaScript to tell it to hide/unhide; the menu toggling is handled by the checkbox input in the CSS. Line 20 can be omitted.

src/lib/components/sections/navbar.svelte
```diff
@@ -1,42 +1,57 @@
...
+						aria-current={segment === slug}>
 						{@html title}
 					</a>
 				</li>
 			{/each}
 		</ul>
 	</div>
 
-	<label for="navToggle" class="menu">
+	<label for="navToggle" class="menu" on:click={() => handleOpen()}>
```
EthanThatOneKid: An `on:click` listener is not needed to open the menu since it is handled by the checkbox input.

EthanThatOneKid (changes requested):
It is exciting to see the menu closing automatically now! I have a few nitpicks, but once those are resolved we can merge this PR.

src/lib/components/sections/navbar.svelte
anhduy1202: Hi, I just changed it

src/lib/components/sections/navbar.svelte
anhduy1202: Done!

src/lib/components/sections/navbar.svelte
anhduy1202: Done :D

EthanThatOneKid (approved):

Issue #143, opened 2021-10-06
Calendar doesn't show recurring events


ghost:
Calendar doesn't show any recurring events via Google Calendar, only the first instance of that event

EthanThatOneKid:
Closed by #160.

PR #155, opened 2021-10-22
[TEST] Recurring Calendar Events

EthanThatOneKid:
## What?

This PR tests if recurring calendar events are able to be rendered on the events carousel on `acmcsuf.com/events`. With some tweaking and refactoring, I think the uploaded implementation works well enough to resolve bug #143.

**Note**: This PR is not meant to be merged into main; this is just to test how this looks in prod.


EthanThatOneKid:
This PR was able to render a recurring event even after its initial event date. The _recurring test event_ was originally scheduled for October 19th at 5:45 PM, which was "last week" at the time of taking the screenshot.

Here is how it would look in prod (shot on my iPad):

IMG_0029

PR #156, opened 2021-10-23
Modify acm-select.svelte to create a custom select dropdown

### LINKED ISSUE #47: Fine-tune `acm-select.svelte` CSS ▶
Currently, the [`acm-select.svelte`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/main/src/components/ui/acm-select.svelte) component looks similar to the current design, but is not identical.

Refer to the [Figma design of the component](https://www.figma.com/file/9cvuO69WgNGuCjf2JGDPfq/ACM-Website-Mockup-Mike-Ploythai?node-id=529%3A919) to fine-tune the CSS in order to match the design:

select-input

Here is the arrow icon as seen on the right side of the component:

Arrow Icon

anhduy1202:
Hi, can you check the design as well as the code for my custom select dropdown? Thanks

### DIFF
--- a/src/lib/components/utils/acm-select.svelte
+++ b/src/lib/components/utils/acm-select.svelte
@@ -1,34 +1,29 @@
 <script lang="ts">
-	export let defaultValue: string | HTMLInputElement = '';
+	export let defaultValue: string = '';
 	export let options: string[] = [];
 
-	let option: HTMLDivElement;
+	let currentValue: string = defaultValue;
 	let dropTitle: HTMLDivElement;
+	let active: boolean = false;
 
-	const setDrop = () => {
-		option.classList.toggle('active');
+	const openDropdown = () => {
+		active = !active;
 	};
 
 	const handleOption = (term) => {
-		dropTitle.innerHTML = term;
-		option.classList.remove('active');
-		defaultValue = term;
+		currentValue = term;
+		defaultValue = currentValue;
+		active = false;
 	};
 </script>
 
 <div class="term" name="school-year">
 	<div class="option-box">
-		<div class="selected" bind:this={dropTitle} on:click={setDrop}>{defaultValue}</div>
-		<div class="option active" bind:this={option}>
+		<div class="selected" bind:this={dropTitle} on:click={openDropdown}>{currentValue}</div>
+		<div class="option" class:active>
 			{#each options as optionValue (optionValue)}
-				<div>
-					<input
-						value={optionValue}
-						on:click={() => handleOption(optionValue)}
-						type="radio"
-						class="radio"
-						id={optionValue} />
-					<label for={optionValue}> {optionValue} </label>
+				<div class="option-choice" on:click={() => handleOption(optionValue)}>
+					{optionValue}
 				</div>
 			{/each}
 		</div>
@@ -45,7 +40,7 @@
 		}
 	}
 
-	label,
+	.option-choice,
 	.selected {
 		color: white;
 	}
@@ -64,7 +59,7 @@
 		visibility: hidden;
 		margin-bottom: 0.5rem;
 		transition: all 300ms;
-		label {
+		.option-choice {
 			cursor: pointer;
 			&:hover {
 				color: transparent;
@@ -82,7 +77,7 @@
 		margin-top: 0.2rem;
 		border-radius: 0 0 6px 6px;
 		transition: all 200ms;
-		label:hover {
+		.option-choice:hover {
 			color: var(--acm-blue);
 		}
 	}
--- a/src/lib/components/utils/acm-select.svelte
+++ b/src/lib/components/utils/acm-select.svelte
@@ -3,10 +3,9 @@
 	export let options: string[] = [];
 
 	let currentValue: string = defaultValue;
-	let dropTitle: HTMLDivElement;
 	let active: boolean = false;
 
-	const openDropdown = () => {
+	const toggleDropdown = () => {
 		active = !active;
 	};
 
@@ -19,7 +18,7 @@
 
 <div class="term" name="school-year">
 	<div class="option-box">
-		<div class="selected" bind:this={dropTitle} on:click={openDropdown}>{currentValue}</div>
+		<div class="selected" on:click={toggleDropdown}>{currentValue}</div>
 		<div class="option" class:active>
 			{#each options as optionValue (optionValue)}
 				<div class="option-choice" on:click={() => handleOption(optionValue)}>
@@ -42,7 +41,7 @@
 
 	.option-choice,
 	.selected {
-		color: white;
+		color: var(--acm-light);
 	}
 	.selected {
 		background-color: var(--acm-dark);
@@ -62,12 +61,9 @@
 		.option-choice {
 			cursor: pointer;
 			&:hover {
-				color: transparent;
+				color: var(--acm-light);
 			}
 		}
-	}
-	.radio {
-		display: none;
 	}
 
 	.option.active {
--- a/src/lib/components/utils/acm-select.svelte
+++ b/src/lib/components/utils/acm-select.svelte
@@ -17,9 +17,9 @@
 </script>
 
 <div class="term" name="school-year">
-	<div class="option-box">
+	<div class="option-box" class:active>
 		<div class="selected" on:click={toggleDropdown}>{currentValue}</div>
-		<div class="option" class:active>
+		<div class="option">
 			{#each options as optionValue (optionValue)}
 				<div class="option-choice" on:click={() => handleOption(optionValue)}>
 					{optionValue}
@@ -47,10 +47,13 @@
 		background-color: var(--acm-dark);
 		padding: 8px 24px;
 		cursor: pointer;
-		border-radius: 6px 6px 0px 0px;
+		border-radius: 6px;
 		&:hover {
 			color: var(--acm-blue);
 		}
+	}
+	.active > .selected {
+		border-radius: 6px 6px 0 0;
 	}
 
 	.option {
@@ -66,7 +69,7 @@
 		}
 	}
 
-	.option.active {
+	.active > .option {
 		visibility: visible;
 		background-color: var(--acm-dark);
 		padding: 8px 24px;
--- a/src/lib/components/utils/acm-select.svelte
+++ b/src/lib/components/utils/acm-select.svelte
@@ -1,37 +1,89 @@
 <script lang="ts">
-	export let defaultValue: string = '';
+	export let defaultValue: string | HTMLInputElement = '';
 	export let options: string[] = [];
+
+	let option: HTMLDivElement;
+	let dropTitle: HTMLDivElement;
+
+	const setDrop = () => {
+		option.classList.toggle('active');
+	};
+
+	const handleOption = (term) => {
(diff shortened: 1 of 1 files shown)


src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,89 @@
 <script lang="ts">
-	export let defaultValue: string = '';
+	export let defaultValue: string | HTMLInputElement = '';
```
EthanThatOneKid: What is the `HTMLInputElement` type used for here?

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,89 @@
...
 	export let options: string[] = [];
+
+	let option: HTMLDivElement;
```
EthanThatOneKid: This element acts as a container for the option elements, we could use a different name like `optionsContainer` or something along those lines.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,89 @@
...
+	let dropTitle: HTMLDivElement;
+
+	const setDrop = () => {
```
EthanThatOneKid: Nitpicking here, but let's rename this function to `openDropdown` to be more clear about what this event handler is called for.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,89 @@
...
+		option.classList.toggle('active');
```
EthanThatOneKid: To open the dropdown menu, we do not need to manually toggle the class, but instead we can set a variable (maybe `active`) to true. Then, using that variable, you can conditionally use the CSS class like so: `<div class:active>...</div>` (more info on <https://svelte.dev/docs#class_name>).

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,89 @@
...
+		<div class="selected" bind:this={dropTitle} on:click={setDrop}>{defaultValue}</div>
+		<div class="option active" bind:this={option}>
+			{#each options as optionValue (optionValue)}
+				<div>
+					<input
+						value={optionValue}
+						on:click={() => handleOption(optionValue)}
+						type="radio"
+						class="radio"
+						id={optionValue} />
```
EthanThatOneKid: This input element is not needed, so we can just use a normal DIV here; preferably the one that starts on line 24. This DIV can steal the `on:click={() => handleOption(optionValue)}` attribute and can use `{optionValue}` as the content of the DIV.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,89 @@
...
+	let dropTitle: HTMLDivElement;
+
+	const setDrop = () => {
+		option.classList.toggle('active');
+	};
+
+	const handleOption = (term) => {
+		dropTitle.innerHTML = term;
+		option.classList.remove('active');
+		defaultValue = term;
```
EthanThatOneKid: The body of this function can be refactored. A good start would be to create a new variable to store the current value (something like `currentValue`). It can be instantiated using the `defaultValue` near the top of the file, but should store the value as it is changed/updated/selected in this function.

EthanThatOneKid (changes requested):
The component itself looks nice! I added some requests and comments for you.

src/lib/components/utils/acm-select.svelte
anhduy1202: oh I forgot to delete it from a previous fix, done btw

src/lib/components/utils/acm-select.svelte
anhduy1202: done :)

src/lib/components/utils/acm-select.svelte
anhduy1202: I set the div class name as optionChoice, is that clear enough? also do you want to be refactor anything from the SCSS or it's good enough?

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: this what I've done, it works so far ( line 23 )
```svelte
class={active ? 'option active' : 'option'}
```

src/lib/components/utils/acm-select.svelte
anhduy1202: can you elaborate more about the use of currentValue? is it storing the value of optionValue?

src/lib/components/utils/acm-select.svelte
anhduy1202: acutally, I was able to delete the optionContainer due to adding the active variables + using class from Svelte

src/lib/components/utils/acm-select.svelte
EthanThatOneKid: That does work technically but there might be an even better way depending on how you look at it.

I think this snippet is nice since it utilizes the `class:*` directive: `<div class="option" class:active>...</div>`.

src/lib/components/utils/acm-select.svelte
EthanThatOneKid: The `currentValue` variable would store the value of `term` when a new term is chosen. You can initialize the variable at the top of the file like so:

```ts
let currentValue = defaultValue;
```

Then, you can use `currentValue` throughout the file to render the currently selected value.

src/lib/components/utils/acm-select.svelte
EthanThatOneKid: That's a good name for the class, but for CSS naming we try to use kabob-case e.g. `option-choice`. 

As for the SCSS, I'll add more comments soon.

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
diamondburned: > Nitpicking here, but let's rename this function to `openDropdown` to be more clear about what this event handler is called for.

It would be more correct to call this function `toggleDropDown`, since it
doesn't actually set it to open when called twice.

EthanThatOneKid: I didn't think of having the button toggle the action variable which is a nice touch by @anhduy1202. With this in mind, I agree that the verb "toggle" does work better in the function name as @diamondburned said.

anhduy1202: @EthanThatOneKid  sounds great! did you check my recent PR? besides from this renaming, any comments on the fix I made? thanks

EthanThatOneKid: I'll reply with any comments I have in a future change request :)

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,84 @@
 <script lang="ts">
 	export let defaultValue: string = '';
 	export let options: string[] = [];
+
+	let currentValue: string = defaultValue;
+	let dropTitle: HTMLDivElement;
+	let active: boolean = false;
+
+	const openDropdown = () => {
```
EthanThatOneKid: > It would be more correct to call this function toggleDropdown, since it doesn't actually set it to open when called twice.

Mentioned in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/156#discussion_r735060397>.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,84 @@
...
+	.option {
 		cursor: pointer;
-		background-color: var(--acm-blue);
+		visibility: hidden;
+		margin-bottom: 0.5rem;
+		transition: all 300ms;
+		.option-choice {
+			cursor: pointer;
+			&:hover {
+				color: transparent;
```
EthanThatOneKid: The color can be set to `var(--acm-light)` since the background color is set to `var(--acm-dark)`.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,84 @@
...
+			}
+		}
+	}
+	.radio {
+		display: none;
 	}
```
EthanThatOneKid: I think the `.radio` selector is currently unused.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,84 @@
 <script lang="ts">
 	export let defaultValue: string = '';
 	export let options: string[] = [];
+
+	let currentValue: string = defaultValue;
+	let dropTitle: HTMLDivElement;
```
EthanThatOneKid: I think the `dropTitle` variable is unused.

src/lib/components/utils/acm-select.svelte
```diff
@@ -1,37 +1,84 @@
...
+	}
+	.selected {
 		background-color: var(--acm-dark);
-		color: var(--acm-light);
-		outline: none;
-		appearance: none;
-		transition: background-color 0.25s ease-in-out;
+		padding: 8px 24px;
+		cursor: pointer;
+		border-radius: 6px 6px 0px 0px;
```
EthanThatOneKid: I was thinking that maybe we could keep all the corners rounded, except when the element is active. One way this can be done is by moving `class:active` from line 23 to the `.option-box` DIV. Then in the CSS, you can do something along the lines of...

```css
.selected {
  border-radius: 6px;
}

.active > .selected {
  border-radius: 6px 6px 0px 0px;
}
```

This change implies that the selector on line 73 must be adjusted as well. From `.option.active` to `.active > .option`.

EthanThatOneKid (changes requested):
I added a few more nitpicks with some more notes on the CSS. Keep up the great work; this is becoming a pretty solid Svelte component!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

src/lib/components/utils/acm-select.svelte
anhduy1202: done!

EthanThatOneKid (approved):
This solid, 88-lined custom select component looks good to me! Feel free to squash and merge your PR whenever you're ready, @anhduy1202.

PR #157, opened 2021-10-24
fixed links for the paths from the homepage 

### LINKED ISSUE #151: Bugged destinations for path links on homepage
### What happened?

Since the SvelteKit migration, the links to the paths from the homepage do not land the user in the correct spots, `#create`, `#algo`, and `#dev`.

> Path section redirect doesn't show path sections properly (ex: clicking algo -> #algo in paths page)

Original: #148

MohamedHabarneh:
When clicking the links for the paths from the homepage, it should put the user in the correct spot.

### DIFF
--- a/src/lib/components/index/acm-paths.svelte
+++ b/src/lib/components/index/acm-paths.svelte
@@ -15,7 +15,7 @@
 
 	<div class="paths-list">
 		{#each acmPaths as { title, slug, picture, color } (slug)}
-			<a class="path-item" href={`/paths#${slug}`}>
+			<a class="path-item" target="_self" href={`/paths#${slug}`}>
 				<img src={picture} alt={`${slug}-logo`} />
 				<p class="size-m headers">
 					acm<span class="brand-em" use:styleProps={{ 'brand-color': color }}>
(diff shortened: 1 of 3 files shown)


EthanThatOneKid (changes requested):
Currently when the user lands on the `/paths#algo` [for example], the user is taken to the correct heading element, but it looks slightly off.

The path icon is slightly cut off meaning that when the user lands on `/paths`, we want there to be an offset such that the path icon is fully visible upon opening either `/paths#algo`, `/paths#create`, or `/paths#dev`.

src/lib/components/index/acm-paths.svelte
```diff
@@ -15,7 +15,7 @@
 
 	<div class="paths-list">
 		{#each acmPaths as { title, slug, picture, color } (slug)}
```
EthanThatOneKid, line 17: The `acmPaths` variable (imported from [`/src/lib/constants/acm-paths.ts`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/5e64fe8e328c33f0f0c71127378068435420ee90/src/lib/constants/acm-paths.ts)) is where we can update to the new hash names `#algo`, `#create`, and `#dev`. The links on the homepage must also be adjusted accordingly.

EthanThatOneKid (changes requested):
We'd also like to minimalize the hashes to `#algo`, `#create`, and `#dev`.

EthanThatOneKid (approved):
Looks good to me! Feel free to squash and merge your PR when you're ready.

PR #163, opened 2021-10-28
Fixed navigation tab bug

dustinchavez1:
- Changed instances of slug to path
- Changed conditional logic for aria-current attribute
- Using new page store from svelte kit


EthanThatOneKid (approved):
Looks great, works great!

PR #178, opened 2021-10-30
Removed unneeded arrow buttons for event carousel

### LINKED ISSUE #159: Unneeded arrow buttons on event carousel ◀▶
### What happened?

When the event carousel holds so few event cards that the contents are not able to scroll horizontally, there should be no buttons to assist in scrolling the carousel. Therefore, when the event carousel in unable to scroll, the arrow buttons on the left and right sides of the carousel should be hidden.

eventcarousel screenshot

Reported by @barrotbake via Discord.

MohamedHabarneh:
When the arrow buttons are unneeded they are hidden and return when the event carousel is scrollable.

Preceding abandoned PR: link-https://github.com/EthanThatOneKid/acmcsuf.com/pull/177. (It included 3 unnecessary files along with the 1 file I wanted to change)

### DIFF
--- a/src/lib/components/events/event-carousel.svelte
+++ b/src/lib/components/events/event-carousel.svelte
@@ -2,6 +2,7 @@
 	import { onMount } from 'svelte';
 	import EventItem from './event-item.svelte';
 	import type { AcmEvent } from '$lib/ical/parse';
+	import { Time } from '$lib/constants/time';
 
 	export let events: AcmEvent[] = [];
 
@@ -13,18 +14,28 @@
 	let hasHorizontalScrollbar = false;
 	let leftButtonEnabled = false;
 	let rightButtonEnabled = false;
+	let scrollTimeId = undefined;
 
 	const scrollTheCarousel = (movementScalar: number, isSmooth: boolean = false) => {
 		carouselRef.scrollBy({
 			left: -movementScalar,
 			behavior: isSmooth ? 'smooth' : 'auto',
 		});
-		const canScrollLeft = carouselRef.scrollLeft > 0;
-		const canScrollRight =
-			carouselRef.scrollWidth - carouselRef.scrollLeft - carouselRef.clientWidth <
-			scrollIncrementDistance;
-		leftButtonEnabled = canScrollLeft;
-		rightButtonEnabled = canScrollRight;
+		debounceTheScroll(() => {
+			const canScrollLeft = carouselRef.scrollLeft > 0;
+			const canScrollRight =
+				carouselRef.scrollWidth - carouselRef.scrollLeft - carouselRef.clientWidth < 1;
+			leftButtonEnabled = canScrollLeft;
+			rightButtonEnabled = canScrollRight;
+		}, Time.Second * 0.75);
+	};
+
+	const debounceTheScroll = (callback: () => void, wait: number) => {
+		callback();
+		if (scrollTimeId !== undefined) {
+			clearTimeout(scrollTimeId);
+		}
+		scrollTimeId = setTimeout(callback, wait);
 	};
 	const scrollOnMouseMove = (event: MouseEvent) => isGrabbing && scrollTheCarousel(event.movementX);
 	const startGrabbing = () => (isGrabbing = true);
@@ -150,7 +161,7 @@
 		-webkit-user-select: none;
 		-ms-user-select: none;
 		user-select: none;
-		visibility: hidden;
+		opacity: 0;
 	}
 
 	.carousel-button:hover {
@@ -169,7 +180,7 @@
 	}
 
 	.enabled {
-		visibility: visible;
+		opacity: 1;
 	}
 	@media (max-width: 1219px) {
 		.event-carousel-container {
--- a/src/lib/components/events/event-carousel.svelte
+++ b/src/lib/components/events/event-carousel.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import EventItem from './event-item.svelte';
 	import type { AcmEvent } from '$lib/ical/parse';
+	import { onMount } from 'svelte';
 
 	export let events: AcmEvent[] = [];
 
@@ -15,6 +16,7 @@
 			left: -movementScalar,
 			behavior: isSmooth ? 'smooth' : 'auto',
 		});
+	var hasHorizontalScrollbar = false;
 	const scrollOnMouseMove = (event: MouseEvent) => isGrabbing && scrollTheCarousel(event.movementX);
 	const startGrabbing = () => (isGrabbing = true);
 	const endGrabbing = () => (isGrabbing = false);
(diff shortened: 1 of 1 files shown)


src/lib/components/events/event-carousel.svelte
```diff
@@ -15,6 +16,7 @@
 			left: -movementScalar,
 			behavior: isSmooth ? 'smooth' : 'auto',
 		});
+	var hasHorizontalScrollbar = false;
```
EthanThatOneKid: This variable is best declared with `let` instead of `var`.

src/lib/components/events/event-carousel.svelte
```diff
@@ -24,13 +26,18 @@
...
 <section>
 	<div class="event-carousel-container">
-		<div bind:this={carouselButtonLeft} class="carousel-button left" on:click={scrollLeft}>
-			&lt;
-		</div>
+		{#if hasHorizontalScrollbar}
+			<div bind:this={carouselButtonLeft} class="carousel-button left" on:click={scrollLeft}>
+				&lt;
+			</div>
+		{/if}
```
EthanThatOneKid: Instead of using Svelte's `{#if}` statement, I think it would make sense to use Svelte's [`class:*` directive](https://svelte.dev/docs#class_name) here. In the DIV that contains the left button, you can actually try adding the attribute below:

```html
<div class:enabled={leftButtonEnabled} />
```

This attribute makes sure that the element's class will be properly toggled depending on the boolean value of `leftButtonEnabled`.

src/lib/components/events/event-carousel.svelte
```diff
@@ -46,9 +53,11 @@
...
 			<div class="event-item-buffer" />
 		</div>
-		<div bind:this={carouselButtonRight} class="carousel-button right" on:click={scrollRight}>
-			&gt;
-		</div>
+		{#if hasHorizontalScrollbar}
+			<div bind:this={carouselButtonRight} class="carousel-button right" on:click={scrollRight}>
+				&gt;
+			</div>
+		{/if}
```
EthanThatOneKid: See the above comment.

EthanThatOneKid (changes requested):
Wow this is a great start! I added some ideas that we discussed on Discord in my review comments.

src/lib/components/events/event-carousel.svelte
MohamedHabarneh: Done.

src/lib/components/events/event-carousel.svelte
MohamedHabarneh: I added the attributed above instead of using the {#if} statement. I'm not sure if you wanted me to create two variables called leftButtonEnabled and rightButtonEnabled, I just used the hasHorizontalScrollbar variable.

src/lib/components/events/event-carousel.svelte
```diff
@@ -24,11 +26,18 @@
...
+		hasHorizontalScrollbar = carouselRef.scrollWidth > carouselRef.clientWidth;
+	});
 </script>
 
 <section>
 	<div class="event-carousel-container">
-		<div bind:this={carouselButtonLeft} class="carousel-button left" on:click={scrollLeft}>
+		<div
+			bind:this={carouselButtonLeft}
+			class:enabled={hasHorizontalScrollbar}
```
EthanThatOneKid: A separate variable `leftButtonEnabled` (and `rightButtonEnabled`) can be initialized and set to false adjacent to `hasHorizontalScrollbar`. Each time `scrollTheCarousel` is called, we should run something that might end up looking like:
```ts
const canScrollLeft = carouselRef.scrollLeft > 0;
const canScrollRight = carouselRef.scrollWidth - carouselRef.scrollLeft - carouselRef.clientWidth < 1; // https://stackoverflow.com/a/42860948

leftButtonEnabled = canScrollLeft;
rightButtonEnabled = canScrollRight;
```

EthanThatOneKid (changes requested):
I added a comment for differentiating between the left and right buttons.

src/lib/components/events/event-carousel.svelte
MohamedHabarneh: I added the leftButtonEnabled and rightButtonEnabled variables in the scrollTheCarousel function so it recalculates each time it's called. The only issue I had was that the first two initial clicks for the right button, scrollLeft is still zero then it begins to increment. So the left button appears after two clicks.

src/lib/components/events/event-carousel.svelte
```diff
@@ -9,12 +10,22 @@
 	let carouselButtonLeft: HTMLDivElement;
 	let carouselButtonRight: HTMLDivElement;
 	let isGrabbing = false;
+	let hasHorizontalScrollbar = false;
+	let leftButtonEnabled = false;
+	let rightButtonEnabled = false;
```
EthanThatOneKid, line 18: Here, let's add this helper function to help us solve the bug with the arrow buttons:

```ts
let scrollTimeoutId = undefined;

const debounceTheScroll = (callback: () => void, wait: number) => {
  callback();
  if (scrollTimeoutId !== undefined) {
    clearTimeout(scrollTimeoutId);
  }
  scrollTimeoutId = setTimeout(callback, wait);
};
```

src/lib/components/events/event-carousel.svelte
```diff
@@ -9,12 +10,22 @@
...
 			left: -movementScalar,
 			behavior: isSmooth ? 'smooth' : 'auto',
 		});
+		const canScrollLeft = carouselRef.scrollLeft > 0;
+		const canScrollRight =
+			carouselRef.scrollWidth - carouselRef.scrollLeft - carouselRef.clientWidth <
+			scrollIncrementDistance;
+		leftButtonEnabled = canScrollLeft;
+		rightButtonEnabled = canScrollRight;
+	};
```
EthanThatOneKid: Let's wrap this code into to helper function we created before.

```ts
debounceTheScroll(() => {
  const canScrollLeft = carouselRef.scrollLeft > 0;
  const canScrollRight =
    carouselRef.scrollWidth - carouselRef.scrollLeft - carouselRef.clientWidth < 1;
  leftButtonEnabled = canScrollLeft;
  rightButtonEnabled = canScrollRight;
}, Time.Second * 0.75);
```

src/lib/components/events/event-carousel.svelte
```diff
@@ -127,6 +150,7 @@
 		-webkit-user-select: none;
 		-ms-user-select: none;
 		user-select: none;
+		visibility: hidden;
```
EthanThatOneKid, line 165: Instead of `visibility: hidden;`, maybe using opacity might be able to help us achieve a good transition effect.

```css
opacity: 0;
```

src/lib/components/events/event-carousel.svelte
```diff
@@ -144,6 +168,9 @@
 		right: 0;
 	}
 
+	.enabled {
+		visibility: visible;
```
EthanThatOneKid, line 185:
```suggestion
              visibility: visible;
		opacity: 1;
```

EthanThatOneKid (changes requested):

src/lib/components/events/event-carousel.svelte
MohamedHabarneh: Done.

src/lib/components/events/event-carousel.svelte
MohamedHabarneh: I replace the visibility with opacity, but the button is still clickable even when it's hidden. Using visibility: hidden fixes that problem so wouldn't it be better to do that?

EthanThatOneKid: We can keep `visibility: hidden`. Adding the `opacity` will enable the transition effect.

EthanThatOneKid (approved):
Great work! Thank you for contributing on this issue.

PR #189, opened 2021-11-03
Fix Spacing component parameters

diamondburned:
This commit fixes the Spacing component's CSS variables being
incorrectly set. They didn't work prior to this change, because of a
regression from PR #184.


EthanThatOneKid (approved):
Thanks for finding this bug!

PR #204, opened 2021-11-07
Fix for website-issue-183

### LINKED ISSUE #183: Strip unwanted backslashes from event titles
### What happened?

When certain characters are included in the event title on Google Calendar (such as commas), a backslash is awkwardly placed as well. To solve this, we can remove all of the backslashes from the title on the backend.

weird-commas

janiljohn:
Added a function called cleanSummary to common.ts which receives a string called summary. The .replace() function is called on summary to replace any \ with an empty string. This function is then exported and used in the parse.ts file.


src/lib/ical/common.ts
```diff
@@ -130,3 +130,10 @@ export const sortByDate = () => {
 export const filterIfPassed = (now: number, offset: number = 0) => {
 	return ({ date }: { date: Date }) => date.valueOf() + offset > now;
 };
+
+
+export const cleanSummary = (summary?: string) => {
+  if(summary === undefined)
+		return "Unnamed Event";
```
EthanThatOneKid: To conform with our style guidelines, I propose we format the code like so:

```suggestion
  if (summary === undefined) return "Unnamed Event";

```

EthanThatOneKid (changes requested):
Well done! I had a minor nitpick on line 136, but I am comfortable merging once that is resolved. Thanks for the contribution, @janiljohn!

src/lib/ical/common.ts
janiljohn: Ah I see. Sounds good to me.

PR #205, opened 2021-11-07
Update common.ts

janiljohn:
Added a return type to the cleanSummary function in the common.ts file.


EthanThatOneKid (commented):
Nice job! Thanks for getting rid of this eslint warning 😄

PR #229, opened 2021-11-11
Created `check_site_changes.sh` script

### LINKED ISSUE #198: Only build when website files are updated
### What happened?

Right now, Vercel builds and spins up a dedicated instance of each pull request we make regardless if our code has anything to do with the website. 

### Solution

To reduce our carbon footprint, it might be a good idea to make a rule that tells Vercel to build if the updated code has a specific file name or is under a specific directory.

EthanThatOneKid:
Created this script to check for any unmerged code that would affect the website output. If there are relevant changes, Vercel will be asked to build.

[image](https://vercel.com/ethanthatonekid/acm-csuf-site/settings/git)

Closing #198 with a new script.


scripts/check_site_changes.sh
```diff
@@ -0,0 +1,7 @@
+# Name: check_site_changes.sh
```
diamondburned, line 3: It'd be better to prepend the following line:

```sh
#!/usr/bin/env bash
```

This ensures that the shell script is evaluated in Bash when it's run directly.

diamondburned (changes requested):

diamondburned:
Has this been tested?

EthanThatOneKid:
> Has this been tested?

It works as intended in #229. 

Since the only change in this PR is a non-website change, Vercel skips building this PR. I tested the script locally to see if it exits correctly when there are changes though and both cases work correctly.

scripts/check_site_changes.sh
EthanThatOneKid: Nice suggestion. Resolved in <https://github.com/EthanThatOneKid/acmcsuf.com/commit/4442ca2fcb661e51cf808fd0ba91aa9fd7d39059>.

diamondburned (approved):

PR #226, opened 2021-11-11
Fix spacing between Board members and semester dropdown menu (Issue #223)

anhduy1202:
This works for me, it doesn't affect spacing of other components I believe

image

PR #233, opened 2021-11-18
Adjusted strange spacing in various places

EthanThatOneKid:
### Background

Found in @mikeploythai's <https://github.com/EthanThatOneKid/acmcsuf.com/pull/226#pullrequestreview-809419818>.

### Before

image

### After

image

### DIFF
--- a/src/routes/about.svelte
+++ b/src/routes/about.svelte
@@ -14,7 +14,7 @@
   };
 </script>
 
-<Spacing --min="100px" --med="175px" --max="200px" />
+<Spacing --min="175px" --med="200px" --max="200px" />
 
 <Header />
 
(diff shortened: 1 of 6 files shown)


ghost (commented):
This fixed the spacing issue for the components, but now the button gap is back on the viewport dimensions of my PC monitor, but not my laptop viewport.  Strange behavior

image

EthanThatOneKid:
> This fixed the spacing issue for the components, but now the button gap is back on the viewport dimensions of my PC monitor, but not my laptop viewport. Strange behavior
> 
> image

I'd say the behavior is strange due to lack of testing. More finetuning is definitely needed in some places. I'll pivot this PR into a general fix to clean up as many of these hidden fixes as we can find. Thanks for noting that.

PR #237, opened 2021-11-28
Fix for issue 211

### LINKED ISSUE #211: Indefinite "Loading..." message on `/events`
### What happened?

The screen says "Loading..." when it should show the empty-events-message instead since there is was an error loading the event data. 

This occurs in 100% of users that have JavaScript disabled on their browser.

### Screenshot

image

janiljohn:
Added a Boolean variable in events.js which check to see if JS is enabled by the user's browser or not. If JS is enabled and there are no events, a message saying "Loading" or "There are no events scheduled!" will be displayed. If JS is disabled and there are no events, nothing will be displayed.

### DIFF
--- a/src/routes/events.svelte
+++ b/src/routes/events.svelte
@@ -52,6 +52,8 @@
         {:else}
           There are no events scheduled!
         {/if}
+      {:else}
+        JS is needed to fetch events
       {/if}
     </p>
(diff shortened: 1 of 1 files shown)


EthanThatOneKid:
Closed #211.

PR #244, opened 2021-12-06
Created initial, read-only newsletters web interface

### LINKED ISSUE #121: Add `/newsletters` to the site 📰
**Note**: "blog" and "newsletter" are interchangeable, but we will be officially going with the term "newsletter".

- [ ] Create a `/blog` endpoint that shows a list of blog posts.
- [ ] Use the `/blog/:slug` endpoint to render full blog posts with [mdsvex](https://mdsvex.com/).
- [x] #125
- [ ] Only show blog posts that are beyond their "release_date" (from preamble).

adityashah6:
- created acmcsuf.com/newsletters.xml
- created web interface for acmcsuf.com/newsletters using acmcsuf.com/newsletters.json
- created web interface for each blog post


src/routes/newsletters/index.svelte
```diff
@@ -32,4 +35,58 @@
   </ul>
 </section>
 
-<pre><code>{JSON.stringify(posts, null, 2)}</code></pre>
+<!-- <pre><code>{JSON.stringify(posts, null, 2)}</code></pre> -->
```
EthanThatOneKid: Oops accidentally left this commented-out code. Please hotfix this.
```suggestion

```

EthanThatOneKid (approved):

EthanThatOneKid:
In order for the `/newsletters` pages to exist in production, a link to it must exist somewhere on the website. For now, maybe we can put a link to the `/newsletters` page on the footer next to the GitHub icon.

src/routes/newsletters/_query.ts
```diff
@@ -38,3 +42,36 @@ export const newslettersQuery = `{
     }
   }
 }`;
+
+/* eslint-disable */
```
diamondburned: Why is `eslint` disabled here?

src/routes/newsletters/[id].svelte
```diff
@@ -1,20 +1,89 @@
...
+    <div class="markdown-body">
+      {@html post.html}
+    </div>
+  </div>
+
+  <Spacing />
+</section>
+
+<style lang="scss">
+  @import '../node_modules/gfm.css/source/gfm.scss';
```
Comment (diamondburned) [inline src/routes/newsletters/[id].svelte]: Why a new dependency?

src/routes/newsletters/index.xml.ts
```diff
@@ -0,0 +1,43 @@
...
+const renderXML = (post: Newsletter): string => {
+  return `<item>
+  <title>${post.title}</title>
+  <description>${post.labels.join(', ')}</description>
+  <link>${website}/newsletters/${post.id}/</link>
+  <pubDate>${new Date(post.lastEdited)}</pubDate>
+  <content:encoded>
+    <div style="margin-top: 50px; font-style: italic;">
+      <strong>
+        <a href="${website}/newsletters/${post.id}">
```
diamondburned: This should be a synopsis of the blog article instead of just having a "Keep reading" link. If it's outside the scope of this PR, I think we should create a new follow-up issue to track it.

diamondburned: Also, is it a good idea to format RSS XML like it's HTML? HTML is much less strict; the generated output might not be conformant.

src/routes/newsletters/[id].svelte
Comment (EthanThatOneKid) [reply src/routes/newsletters/[id].svelte]: GFM.css (GitHub-flavored markdown) is a library that mimics the styles used for READMEs on github.com. The idea was that we'd use GFM so the newsletters would look similar on both GitHub and on acmcsuf.com/newsletters. GitHub has a [_Preview_ feature](https://github.blog/changelog/2021-11-17-preview-the-markdown-rendering-of-gists/) which is also helpful for drafting our newsletters. With that being said, GFM is merely a starting point. The end goal is to use our own styles, allowing the newsletter content to blend in seamlessly with the rest of the site. However, for the scope of this PR, GFM is good.

src/routes/newsletters/index.xml.ts
EthanThatOneKid: HTML is less strict indeed, so instead, we could use a different approach than using a string template. I have not used this library personally, but [node-rss](https://github.com/dylang/node-rss) seems to be a possible choice for this purpose. The node-rss repository has not been updated in over 5 years, but that might be fine since RSS feeds are not expected to change anytime soon AFAIK.

src/routes/newsletters/_query.ts
EthanThatOneKid: Good question. There was likely an eslint error that was not going away easily. Any unnecessary eslint comments (including the one you've mentioned) should be removed before this PR is fully approved/merged.

ghost (approved):
LGTM, will get started on official design

diamondburned (approved):
Giving the approval, though I still think we don't actually need `gfm.css`. That can probably be removed in the future once we can polish up our base stylesheet.

ghost (approved):
LGTM

PR #248, opened 2021-12-21
Added Size Limit for `node_modules`

### LINKED ISSUE #247: Add a size limit for our NPM dependencies
### What's your idea?

To make sure we aren’t creating an NPM dependency black hole that takes forever to install on slow connections, we can set a size threshold for our NPM dependencies. 

Related: [ai/size-limit](https://github.com/ai/size-limit#js-applications)

EthanThatOneKid:
I used this NPM package, [ai/size-limit](https://github.com/ai/size-limit), to help with calculating the size of our `node_modules` directory after a clean `npm i`. The size of `fix/247`'s `node_modules` was clocked at 39.91 MB gzipped (according to [run 4835296416](https://github.com/EthanThatOneKid/acmcsuf.com/runs/4835296416?check_suite_focus=true#step:8:23)), so I set the threshold to 50 MB as a starting point.

As we change our `package.json`, we should also contemplate and refer to the threshold in our `package.json`. If anything, we'd like to eventually _lower_ the threshold as time goes on.

Resolves #247.

### DIFF
--- a/.github/workflows/check_website.yaml
+++ b/.github/workflows/check_website.yaml
@@ -45,5 +45,9 @@ jobs:
       - name: Check
         run: npm run check
 
+      # info: https://etok.codes/acmcsuf.com/blob/main/CONTRIBUTING.md#npm-bloat
+      - name: Test Dependency Size
+        run: npm run test:size
+
       - name: Build
         run: npm run build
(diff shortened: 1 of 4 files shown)


diamondburned (approved):
I'm giving the approval for this PR, though it's not the best idea that the limiting metric is the overall modules size and not tree size or whatnot. Something is better than nothing, though.

PR #280, opened 2022-02-03
Make EventCarousel a vertical list

### LINKED ISSUE #278: events carousel button doesnt snap to events
### What happened?

https://user-images.githubusercontent.com/8981287/152116007-783e8f67-ce70-4617-be8c-87d4df1e788c.mp4

diamondburned:
This commit removes most of `EventItem`'s JavaScript for link and scroll handling, and replaces it with a traditional vertical list with an anchor `<a>` tag.

Fixes #278 by a complete refactor.


src/lib/components/events/event-item.svelte
```diff
@@ -1,135 +1,208 @@
...
+  <!-- Workaround for the top panel covering the event card's anchor. -->
+  <div class="anchor" id={info.slug} bind:this={anchor} />
+  <details class="event-card" bind:this={details}>
+    <summary class="event-body">
+      <div class="event-name">
+        <h2 class="headers">
+          <a href="#{info.slug}">
+            {info.summary}
+          </a>
+        </h2>
```
EthanThatOneKid: The inner link can be left out from this `<h2>` element since it interferes with the opening of the `<details>` element when the visitor goes to click the title.

src/lib/components/events/event-carousel.svelte
```diff
@@ -1,108 +1,15 @@
...
-  }
-
-  onMount(() => {
-    hasHorizontalScrollbar = carouselRef.scrollWidth > carouselRef.clientWidth;
-    rightButtonEnabled = hasHorizontalScrollbar;
-  });
 </script>
 
 <section>
   <div class="event-carousel-container">
```
EthanThatOneKid: Nit: This component is no longer a carousel, so it does not feel right to call it that anymore. Something simple like `event-list`?

EthanThatOneKid (changes requested):
I left one note, but this vertical event list looks awesome so far! Can't wait to see it in production.

src/lib/components/events/event-item.svelte
diamondburned: There should be another way to obtain the anchor link if we want to remove the `<a>` tag. Perhaps a copy button inside.

EthanThatOneKid: Good call! Here is a new bug for that: #281.

src/lib/components/events/event-item.svelte
```diff
@@ -173,6 +173,7 @@
     margin: 0;
     padding: 0;
     padding: 12px 24px;
+    width: auto;
```
diamondburned: Did this work?

src/lib/components/events/event-item.svelte
```diff
@@ -174,6 +174,7 @@
     padding: 0;
     padding: 12px 24px;
     width: auto;
+    text-align: center;
```
diamondburned, line 179: Did this work as well?

src/lib/components/events/event-item.svelte
```diff
@@ -29,7 +29,7 @@
     <summary class="event-body">
       <div class="event-name">
         <h2 class="headers">
-          <a href="#{info.slug}">
+          <a href="#{info.slug}" on:click={() => (details.open = !details.open)}>
```
diamondburned: Just remove the `<a>` tag entirely. This JS isn't useful and introduces different behaviors with and without JS.

src/lib/components/events/event-item.svelte
EthanThatOneKid: Removed the `<a>` tag in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/280/commits/219dbbd436a8ca7433a1313a003ff2795d7f2879>.

src/lib/components/events/event-item.svelte
EthanThatOneKid: This centered the "Join" text as expected.

src/lib/components/events/event-item.svelte
EthanThatOneKid: This didn't do much of anything, so I reverted it back in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/280/commits/8cf969c6b0ae3ad65577ebb9c168ddf62242fa3e>.

src/lib/components/events/event-carousel.svelte
EthanThatOneKid: Renamed in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/280/commits/c47f784a8d9e9780bfcd14411fd6d7b2ccee5fe0>.

src/lib/components/events/event-item.svelte
EthanThatOneKid: Got it. Removed in <https://github.com/EthanThatOneKid/acmcsuf.com/commit/219dbbd436a8ca7433a1313a003ff2795d7f2879>.

src/lib/components/events/event-item.svelte
```diff
@@ -119,12 +117,11 @@
...
   }
 
-  .event-body a {
-    text-decoration: none;
+  .event-body .event-name {
     transition: all 0.15s ease-in-out;
   }
 
-  .event-body a:hover {
+  .event-body .event-name:hover {
```
diamondburned: I think we should do `.event-body:hover .event-name` so the event name changes color when the card is hovered on top. It just makes more sense, since the fake link appearance doesn't actually behave like a link anyway.

EthanThatOneKid: I agree that makes more sense. I updated the CSS a little in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/280/commits/60fea4f7cf182160593af2cdf97c3cd72c39557b> to achieve this.

EthanThatOneKid (approved):

PR #288, opened 2022-02-11
Events: Migrated from CSR to SSR in SvelteKit

### LINKED ISSUE #287: Events: Migrate from CSR to SSR
**tl;dr**: Refactor how `acmcsuf.com/events` renders the event list. Make sure the HTML for the cards is generated on the server, instead of retroactively on the client's side.

Currently, `acmcsuf.com/events` fetches the event data from `acmcsuf.com/events.json` on the client after the page loads. This forced us to display a "loading..." dialog during the moment that it takes to load the event data. For our purposes, it is better to bake the event data into the HTML before it is sent to the client so that the user does not have to wait for the event data to be fetched. 

To resolve this bug, the event list items must be visible from `acmcsuf.com/events`'s page source.

EthanThatOneKid:
I migrated the client-side rendered implementation of our event list to a server-side rendering technique. This skips the loading animation and allows visitors to see the events without having JavaScript enabled.

Resolves #287.


EthanThatOneKid:
Here are some screenshots showing how CSR looks (before) and how SSR looks (after) in the raw HTML.

Before:

image

After: 

image

adityashah6 (approved):

jaasonw (approved):

PR #293, opened 2022-02-12
Events: Join from new tab

### LINKED ISSUE #292: Events: Clicking "Join" navigates to 404
### What happened?

Currently when the visitor clicks a "Join" button for an event that links to another `acmcsuf.com`-hosted page, they encounter the 404 error page. We want the visitor to be redirected to the appropriate page.

EthanThatOneKid:
Upon clicking a "Join" button, the join link is opened in a new tab using `target="_blank"`.

Resolves #292.

### DIFF
--- a/src/lib/components/events/event-item.svelte
+++ b/src/lib/components/events/event-item.svelte
@@ -45,7 +45,12 @@
           {#if isRecurring}(recurring){/if}
         </time>
       </p>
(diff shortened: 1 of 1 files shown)


jaasonw (approved):
Can confirm it works, technically `noopener noreferrer` is default behavior in all modern browsers now but it doesn't hurt to be safe

https://chromestatus.com/feature/6140064063029248

Issue #310, opened 2022-03-02
Events: <EventItem /> enhancements tracker


EthanThatOneKid:
#### Actionbar Enhancement

- [x] <https://github.com/EthanThatOneKid/acmcsuf.com/issues/281>
- [x] #312
- [x] #317
- [x] #327

#### Related Bugs/Enhancements

- [x] <https://github.com/EthanThatOneKid/acmcsuf.com/issues/309>
- [x] <https://github.com/EthanThatOneKid/acmcsuf.com/issues/313>
- [x] #330
- [x] #334
- [x] #346 
- [x] #422

#### Future enhancements

- [ ] #181
- [x] #339

EthanThatOneKid:
This umbrella has been ready to resolve since [v3.0](https://github.com/EthanThatOneKid/acmcsuf.com/releases/tag/v3.0).

PR #314, opened 2022-03-03
Fixed button hover effect

### LINKED ISSUE #313: Join button doesn't have hover effect
### What happened?

image

We want to make sure that the summary is still selectable when the hovering effect of button is added

EthanThatOneKid:
# Screenshot

Demo gif

Fixed #313.
Related: #310.


anhduy1202:
LGTM!

PR #319, opened 2022-03-04
Added `npm run all` 👨‍💻

EthanThatOneKid:
The following command is used to check everything you'd normally want to check before committing:

```bash
npm run all
```

Hopefully this prevents future me from accidentally deploying broken branches to Vercel.


jaasonw:
fun fact: u could use a git hook to do things automatically before or after a commit n shit

PR #328, opened 2022-03-11
Swap "share" button for "copy link" button if `navigator.share` is supported

### LINKED ISSUE #327: Events: Create share action for each event item
Use the [`Navigator.share`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share) API to progressively enhance our events page with shareability.

EthanThatOneKid:
If the visitor's device supports the [`Navigator.share`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share) API, then the "copy link" button is swapped out for the "share" button since their functionalities are overlapping. The "copy link" button can be seen as a fallback for visitors who are visiting on browsers that do not happen to support sharing.

Resolves #327.


mle225 (approved):
Looks good, approved 👍

jaasonw:
>https://user-images.githubusercontent.com/8981287/158007331-1db60bdc-d573-4553-81bd-2fc91ddaa080.mp4

EthanThatOneKid:
I will be closing this PR since the web's `navigator.share` API seems to not work as we expect it to uniformly on all devices. Minor inconsistencies causes this feature to be difficult to maintain even if we get it working. For these reasons, I am closing the PR and will close the issue once it has gone stale.

PR #329, opened 2022-03-12
Show set amount of events on `acmcsuf.com/events` when DEBUG=1

EthanThatOneKid:
I introduced a new environment variable `VITE_DEBUG` to make it easier to see certain parts of the website from a debugging point of view. For example, it would be easier for us to debug the acmcsuf.com/events page if there were always event item components on it to test with.


EthanThatOneKid:
This works as seen on the preview URL provided by Vercel: <https://acm-csuf-site-git-ops-debug-1-ethanthatonekid.vercel.app/events>.

image

anguzz (approved):

PR #333, opened 2022-03-12
Resolved annoying 'css-unused-selector' warning messages

### LINKED ISSUE #332: Lots of 'css-unused-selector' warning messages
### What happened?

The only places on the site where we should be overfitting CSS class declarations is on pages with unexpected HTML content, such as the blog post pages. However, we see a many unused CSS selectors coming from our [`theme.scss`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/a32329aa702ab89d919202fa745f803ef83bb7ec/static/theme.scss) file that contains helper classes for sizing using a third-party library, [`rfs`](https://github.com/twbs/rfs#readme). These helper classes can be rewritten on each component where it is needed since it merely requires the SCSS mixin from the 3rd party `rfs` library.

### Screenshot

image

Suggested by @MohamedHabarneh during the first **acmDev** meeting of our Spring 2022 sprint.

EthanThatOneKid:
- Deleted `theme.scss` file
- Added size variables to `global.css`
- Replaced RFS library with CSS `calc()`-ulations in `global.css`
- Removed `gfm.css` (GitHub Flavored Markdown) CSS library

Resolves #332.


ghost (approved):
rfs package doesn't resize some headers to become smaller on mobile breakpoint like before, but tbh i kinda like it. approved

image

MohamedHabarneh (approved):

EthanThatOneKid:
> rfs package doesn't resize some headers to become smaller on mobile breakpoint like before, but tbh i kinda like it. approved

I added the RFS-generated values to our `global.css` since we don't need to recalculate them.

<https://github.com/EthanThatOneKid/acmcsuf.com/blob/2bbd0da2297d85b939c98414292cdd082f00dc69/static/global.css#L115-L123>

### Updated Screenshot

image

ghost (changes requested):
Join button text should be a big larger for desktop
image

Hover on join button is different now. The background color should be changing to acmBlue, not the text color.

PR #402, opened 2022-03-27
Index page adjustments

ghost:
#### CHANGELOG
- Changed font macros for paragraphs from `size-xs` to `size-sm`
- Breakpoint is set to 900px
- Images use `clamp()` to scale dynamically
- `AcmButton` re-written to use `--button-bg` and `--button-hover` for different styles in light/dark themes
- Modified wording in some paragraphs

This is change #3 of my big update to prepare for next semester's codebase.

Mods to AcmButton (and AcmSelect in another PR) fixes #347.

PR #415, opened 2022-03-31
Mike's index page

### LINKED ISSUE #347: Button hover effect
### What happened?

i recently heard that the plan is to change text color on button hover rather than background-color of the button on button hover. i disagree with this because it's not a good UX choice due to the colors we use.

as it currently stands, the colors we use in our design language is bad for accessibility on dark backgrounds (one of the reasons why I strayed away from dark mode at the early stages of the website) due to how saturated the colors are. relating to this, acmBlue does not look good on acmDark because both colors are too dark. this is why I designed the buttons to change the background color to acmBlue on hover.

i recommend that for buttons, background-color change should be the hover effect.

jaasonw:
> ### PROPOSAL
> 
> > Minimal cleanup. Reworded some stuff.
> 
> #### CHANGELOG
> 
>     * Changed font macros for paragraphs from `size-xs` to `size-sm`
> 
>     * Breakpoint is set to 900px
> 
>     * Images use `clamp()` to scale dynamically
> 
>     * `AcmButton` re-written to use `--button-bg` and `--button-hover` for different styles in light/dark themes
> 
>     * Modified wording in some paragraphs
> 
> 
> This is change #3 of my big update to prepare for next semester's codebase.
> 
> Mods to AcmButton (and AcmSelect in another PR) fixes #347.

https://github.com/EthanThatOneKid/acmcsuf.com/pull/402


EthanThatOneKid:
> [12:57 PM] moike: Okok go ahead and close 426, 415, 414, 413, and 406

PR #431, opened 2022-04-15
Added replaceHtmlWithExternalLinks function to utils.ts

### LINKED ISSUE #346: can u make all links in events open in a new tab
### What's your idea?

[10:54 AM] jasonw: can u make all links in events open in a new tab pls thx
[10:55 AM] EthanThatOneKid: Good idea
[10:56 AM] EthanThatOneKid: Which links were you talking about
[10:56 AM] jasonw: all of them
[10:56 AM] jasonw: clicking anything on that page should never make me leave

janiljohn:
Fix for issue 346. Added a function named replaceHtmlWithExternalLinks in utils.ts that parses through any anchor tags and sets the target attribute to _blank. This will open up any links found within the description of an event in a new tab. Additional tests were also added to utils.test.ts that checked for the correctness of this function.

### DIFF
--- a/src/lib/ical/utils.test.ts
+++ b/src/lib/ical/utils.test.ts
@@ -2,7 +2,14 @@ import { assert, test, expect } from 'vitest';
 import { Temporal } from '@js-temporal/polyfill';
 import { readFileSync } from 'fs';
 import { parse } from './parse';
-import { makeEventSlug, parseDescription, makeEventLink, wrapText, walkICAL } from './utils';
+import {
+  makeEventSlug,
+  parseDescription,
+  makeEventLink,
+  wrapText,
+  walkICAL,
+  replaceHtmlWithExternalLinks,
+} from './utils';
 
 test('parses sample ICAL payload', () => {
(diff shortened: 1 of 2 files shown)


src/lib/ical/utils.test.ts
```diff
@@ -69,3 +76,32 @@ test('wraps long text into lines broken at column 100 3 times', () => {
...
+  example`);
+});
+
+test('additional replaces HTML with external links tests', () => {
+  const actual =
+    replaceHtmlWithExternalLinks(`<a title="example" target="_self"  href="https://example.com/">
+  <a title="example" target="_self"href="https://example.com/">
+  <a target="_self" title="example" href="https://example.com/">
+  <a title="example" target="_self" target="poggers" target="not poggers"  href="https://example.com/">
+  example`);
```
EthanThatOneKid, line 104: What if there are any tags that begin with `<a`? For example, does the input `<article href="https://example.com/">example</article>` remain unchanged as expected?

EthanThatOneKid (approved):
Clean implementation and thorough tests; LGTM!

PR #432, opened 2022-04-15
Updates dependencies

### LINKED ISSUE #411: Automate NPM dependency upgrades
### What's your idea?

It is important to make sure that the project is not falling behind in terms of NPM dependencies. It is recommended online to update our dependencies' versions every month or two. We can create a GitHub Workflow file that can handle this by generating a PR with the updated versions. This would allow us to immediately see if there are any breaking changes with any of our dependencies while staying up-to-date with their latest features.

### Related

> I am thinking we create a new GitHub Workflow file in a future PR that automatically generates a PR updating our dependencies to their latest versions at the beginning of each month. Which command(s) did you run to update the dependencies?

<https://github.com/EthanThatOneKid/acmcsuf.com/pull/409#pullrequestreview-926943937>

### Resources

- [Quickstart for GitHub Actions](https://docs.github.com/en/actions/quickstart)
- [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions)

MohamedHabarneh:
This workflow will create a pr request to update the dependencies at the first of every month at midnight.

### DIFF
--- a/.github/workflows/update_deps.yaml
+++ b/.github/workflows/update_deps.yaml
@@ -1,0 +1,25 @@
+name: Update Dependencies
+on:
+  schedule:
+    - cron: '0 0 1 * *'
+
+jobs:
+  update-dep:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+      - name: Update dependencies
+        run: |
+          npx -p npm-check-updates ncu -u
+          npm install
+      - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


EthanThatOneKid (approved):
**Note**: This PR to add a new workflow is expected to run for the first time on Sunday, May 1st for the first time.

**Note**: We similarly create a new PR to automate officer data in [`.github/workflows/update_officer.yaml`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/0f3c185899984c108ef1fef6d3367ad197d8b7d8/.github/workflows/update_officer.yaml#L45-L55).

LGTM!

PR #439, opened 2022-04-18
Updated page title

### LINKED ISSUE #296: Page title is incorrect after navigating
### What happened?

> ...when u go to a 404 and click back
> the title doesnt change back

_reported by @jaasonw on Discord._

NLTN:
**What caused the bug**
Only the 404 error page has a title, whereas other pages don't. So, after being redirected to the 404 page, if the user clicks back or goes to other pages, the title doesn't change back to normal. It keeps saying "ACM at CSUF / 404" forever, no matter where the user goes.

**What I changed to fix this issue** 

- I added a page tile to the header of the following pages: About, Events, Node Buds, and Paths.

- Also, I removed the title from the file "__layout.svelte" because every page has its own title from now on.


jaasonw (approved):
Well first thing's first, thank you for opening your first pull request with us, any help we get is greatly appreciated : )
Secondly, even though it's a small change, you should always write a brief description of what you changed and why (ex: to fix issue #xyz or something along those lines)
Third, is while it looks like this pull request was originally intended to fix https://github.com/EthanThatOneKid/acmcsuf.com/issues/296, it doesn't look like I can reproduce the bug on my local copy of the main branch right now, so I guess it's technically fixed?
In which case, this looks alright to me as long as other people are on board with using | as a divider

NLTN:
I just updated the description what I did.

One more thing I want to add. This is just a technical fix to solve the problem. Not about changing the naming conventions of the page titles.

jaasonw:
image

Ohhhhhhh so thats how you reproduce it, in which case, the bug dies still exist in prod. Good catch and ty for updatng the description

PR #449, opened 2022-04-21
Rearranged board members

ghost:
As per karni's request. Final change to tiers.json


jaasonw:
Can we bump historian up some ranks too

ghost:
> Can we bump historian up some ranks too

no

jaasonw:
> > Can we bump historian up some ranks too
> 
> no

D:<

k4rni:
Sorry for the late response but we're nearing the end of this semester and I feel like there's no point in updating the tiers at this very moment when the entire list will be changing so I'm just gonna deny this and then we can figure it all out again next semester.

jaasonw:
F in the chat

PR #450, opened 2022-04-23
Added QR Code feature 

### LINKED ISSUE #140: Create QR code endpoint 🔲
For each link in the [links library](https://github.com/EthanThatOneKid/acmcsuf.com/blob/6fcdc3db3bde819833667cf961c306c7ec2186d9/src/lib/links.ts), there should be an endpoint like `acmcsuf.com/qr/discord.png` that sends a PNG of the link [generated as a QR code](https://github.com/SumiMakito/Awesome-qr.js).

QR code mock-up for **acmCreate** links:

LinkedIn_Event

With SvelteKit, the entrypoint for this service should reside in a file (`/src/routes/[link].png.ts`) that exports an HTTP handler which generates/serves these QR codes on command.

anguzz:
Added endpoint which sends a SVG of the link generated as QR code 
 Resolves #140


package.json
```diff
@@ -27,16 +27,20 @@
   "dependencies": {
     "@js-temporal/polyfill": "^0.4.0",
     "html-to-text": "^8.1.0",
+    "qr-code-styling": "^1.6.0-rc.1",
+    "qrcode": "^1.5.0",
     "rrule": "^2.6.9",
-    "rss": "^1.2.2"
+    "rss": "^1.2.2",
+    "svelte-qrcode": "^1.0.0"
```
EthanThatOneKid: Based on the rest of the changes in your PR, I can tell we ended up just using the `qrcode` package. So, you can go ahead and run `npm uninstall qr-code-styling svelte-qrcode` since they are unused dependencies.

src/lib/constants/utils.ts
```diff
@@ -0,0 +1,4 @@
+export function clean(link: string): string {
```
EthanThatOneKid: I was feeling a little weird about putting utility functions under a `/constants/` directory, so perhaps we can move some code into a new file structure:

```
src/
├─ lib/
│  ├─ qr/
│  │  ├─ utils.ts
│  │  ├─ utils.test.ts
```

In the proposed file structure, you can put the code for the `clean` function into `/src/lib/qr/utils.ts`, but we can probably be more specific by naming it something like `normalizeLinkName`. Along with this function, we also need to move the `utils.test.ts` file into the proposed `/src/lib/qr/` directory. This way, you can now delete `/src/lib/constants/utils.ts` and `/src/lib/constants/utils.test.ts`.

We can also move the code defining the `genQRCodeSVG` function (from `src/routes/[link=link].svg.ts`) into the proposed `utils.ts` file to organize it better.

```ts
// File: /src/lib/qr/utils.ts

// This function is meant to clean links against data from `$lib/constants/links`.
export function normalizeLinkName(name: string): string {
  return name.replace(/[^a-zA-Z0-9-]/g, '').toLowerCase();
}

export function genQRCodeSvg(url: string): Promise<string> {
  return new Promise((resolve, reject) => {
    QRCode.toString(url, { type: 'svg' }, (err, svg) => {
      if (err) reject(err);
      resolve(svg);
    });
  });
}
```

EthanThatOneKid (changes requested):
Your PR adequately solves the issue at hand very well! I thought of a minor refactor having to do with me having second thoughts on the `/src/constants/utils.ts` file. I am excited to see this feature launch soon!

src/lib/constants/utils.ts
anguzz: Okay! In the new commit I restructured the functions and re named the clean link.

EthanThatOneKid (approved):
SHEEESH 😎 This PR looks good to me!

Issue #422, opened 2022-04-08
Recent Events: Create secondary list to display `ACM_PINNED=true` events


EthanThatOneKid:
To create this secondary list of pinned events items, we must filter only `AcmEvent`s where `isPinned` is true. To check if an Google Calendar event is intended to be pinned, we must check for an `ACM_PINNED` variable in the description. If that variable’s value is _truthy_, the event `isPinned`.

EthanThatOneKid:
We already check to see if a string is truthy in 

<https://github.com/EthanThatOneKid/acmcsuf.com/blob/4f46f7c1b4843f9535c01f19ac3e715d18b01027/src/lib/constants/index.ts#L5-L6>

So, we can abstract that into a new utility function to be used in place above and again in [`makeAcmEvent`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/4f46f7c1b4843f9535c01f19ac3e715d18b01027/src/lib/ical/utils.ts#L281) to check if an event is pinned.

EthanThatOneKid:
"Recent events" are not necessary to be showcased on the front events page of our website. Perhaps it would be more appropriate to dedicate a new page to recent events, or adding blog posts (with a `past-event` label) to recount the event.

Despite technically resolving this tracked issue in #477, I will close this issue since we are no longer interested in having this feature located on `/events`.

PR #455, opened 2022-04-28
Birth of `parseBool()`

### LINKED ISSUE #476: Utility function for finding truthiness of a string
### What's your idea?

There are several places in this repository that would take advantage of a convenient utility function that checks if a given string is truthy.

mle225:
- Created `src/lib/common/utils.ts` for general purpose utility functions 
- Moved `isStringTruthy` there
- Modified DEBUG constant to use the new `isStringTruthy` function 
- Modified `acmEvent` interface to include `isPinned : boolean`  attribute


src/lib/common/utils.ts
```diff
@@ -0,0 +1,7 @@
+export function isStringTruthy(str: string): boolean {
```
EthanThatOneKid: Since `import.meta.env[string]` can be type of `string` or `undefined`, we must account for the case where an environment variable is `undefined` (or even `null` for fun).  By default, we should set `defaultValue` to `false`.

Example 1:
```ts
// When VITE_DEBUG is left undefined, we want debug mode to be turned off.
console.log(isStringTruthy(import.meta.env.VITE_DEBUG, false));

// Output: false
```
Example 2:
```ts
// .env
// VITE_DEBUG=1

// We only want debug mode to be on when it is set to a truthy value in the .env file
// or Vercel build previews.
console.log(isStringTruthy(import.meta.env.VITE_DEBUG, false));

// Output: true
``` 

To get started on improving the usage for catching `undefined` and `null` edge cases,  let's start by refactoring the function signature.

```suggestion
export function isStringTruthy(payload?: string | null, defaultValue = false): boolean {
```

**Note**: The optional argument syntax (`?`) is used in the above example for the `payload` argument to signify that it can be `undefined`. It is shorthand for otherwise saying `(payload: string | undefined | null, ...)`.

src/lib/common/utils.ts
```diff
@@ -0,0 +1,7 @@
...
+  if (str == null) {
+    return false;
+  }
```
EthanThatOneKid: To test the _truthiness_ of the `payload` argument, let's allow JavaScript to help decide for us. JavaScript's [rules on _truthiness_](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) do most of what we require, most importantly `null`, `undefined`, and `''` (empty strings) are considered [_falsy_](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).
```suggestion
  if (!payload) return false;
```

**Note**: The `!` (negation) operator is used to determine the _falsiness_ of the `payload` argument.

src/lib/common/utils.ts
```diff
@@ -0,0 +1,7 @@
...
+
+  return !['false', '0', ''].includes(str.toLowerCase());
```
EthanThatOneKid: Now that the `undefined`s, `null`s, and empty strings have been ruled out by the above if-statement, we only need to check if the payload matches `"false"` and `"0"` (since they would otherwise be classified as truthy).
```suggestion
  return !['0', 'false'].includes(payload.trim().toLowerCase());
```

EthanThatOneKid (commented):
Thank you for publishing your progress on #422! I left a few comments with my suggestions for techniques writing the `isStringTruthy` utility function 😎

**Reminder**: Once your PR is no longer a WIP, update its name to accurately reflect the changes you will have made.

src/lib/common/utils.ts
mle225: Thanks for your input Ethan, really appreciate it! In this case, if we have checked for truthiness with `if (!payload)` above, would we still need to check for `0` and `false` ? If believe the check also rules out 0, false and other falsy values in JS.

src/lib/common/utils.ts
mle225: Great suggestion!

src/lib/common/utils.ts
EthanThatOneKid: Of course; thank you for contributing!

The `if (!payload)` above only catches the `0` literal and `false` literal values. If `payload` happens to be a string value (`"0"` or `"false"`) we need to manually check `"0"` and `"false"` to determine how we want to discern the truthiness of `payload`.

I hope I am understanding your question correctly 😄

mle225: Yes this makes total sense, I will go ahead and resolve this now, thanks!

src/lib/common/utils.test.ts
```diff
@@ -0,0 +1,26 @@
...
+for (const input of TRUTHY_INPUT_DATA) {
+  test(`correctly determines input value '${input}' as truthy`, () => {
+    expect(utils.determineTruthyString(input)).toBe(true);
+  });
+}
+
+const FALSY_INPUT_DATA = [
+  'FALSE',
+  '',
+  ``,
```
EthanThatOneKid: Nit: This test case seems to be a duplicate.

src/lib/common/utils.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  null,
+  undefined,
+  String(false),
+  String(NaN),
+  String(0),
+  String(-0),
```
EthanThatOneKid: Nit: This test case seems to be a duplicate.

EthanThatOneKid (commented):
I noticed some duplicate input data being used ([here](https://github.com/EthanThatOneKid/acmcsuf.com/runs/6504996947?check_suite_focus=true#step:8:19) and [here](https://github.com/EthanThatOneKid/acmcsuf.com/runs/6504996947?check_suite_focus=true#step:8:25)) which can be removed.

This PR currently introduces the ~~`determineTruthyString`~~ `parseBool` utility function which is used to resolve #422. Even though the function is used in our solution to #422, I think it would make sense to keep this PR more concise; limiting this PR to the changes related to the ~~`determineTruthyString`~~ `parseBool` function in the effort to instead resolve #476.

src/lib/common/utils.test.ts
EthanThatOneKid: Resolved by <https://github.com/EthanThatOneKid/acmcsuf.com/pull/455/commits/ec895ff296a0ba929d68b7864dacb287bc363b81>.

src/lib/common/utils.test.ts
EthanThatOneKid: Resolved by <https://github.com/EthanThatOneKid/acmcsuf.com/pull/455/commits/ec895ff296a0ba929d68b7864dacb287bc363b81>.

src/lib/common/utils.ts
```diff
@@ -0,0 +1,10 @@
+export function parseBool(payload?: string | null, defaultValue = false): boolean {
```
jaasonw: document what this function does because I couldn't immediately tell from the name without reading the implementation

src/lib/common/utils.ts
```diff
@@ -0,0 +1,10 @@
...
+  if (!payload) return defaultValue;
+
+  payload = payload.trim().toLowerCase();
+
+  // unlikely NaN will be passed, but included to be safe
+  const stringsOtherwiseTruthy = [0, false, NaN].map((v) => String(v).toLowerCase());
```
jaasonw: image

why not just precompute this?

jaasonw: feel like this whole function can be written in 1 line like this but not sure if we sacrifice too much readability for this
```ts
export function parseBool(payload?: string | null, defaultValue = false): boolean {
  return ![ '0', 'false', 'nan' ].includes(payload?.trim()?.toLowerCase()) || defaultValue;
}

Comment (jaasonw) [reply src/lib/common/utils.ts]: i also didnt test this yet it's just a quick n dirty proof of concept i did in notepad

Comment (jaasonw) [reply src/lib/common/utils.ts]: image

Update: I tested it

```ts
export function parseBool(payload?: string | null, defaultValue = false): boolean {
  if (!payload) return defaultValue;
  return !['0', 'false', 'nan'].includes(payload?.trim()?.toLowerCase());
}
```

jaasonw (changes requested):

src/lib/common/utils.test.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from 'vitest';
+import * as utils from './utils';
+
+const TRUTHY_INPUT_DATA = ['TRUE', String(true), String(1)];
```
jaasonw: also recommend adding an arbitrary "valid string" to this test case

EthanThatOneKid: Good observation. The `TRUTHY_INPUT_DATA` should definitely include a case for "any non-falsy string" since any string passed to `parseBool` that isn't falsy should return true. For example:
```ts
parseBool("example non-falsy string"); // Output: true
```

src/lib/common/utils.ts
EthanThatOneKid: Since the `parseBool()` function is expected to run many times on the server, perhaps precomputing could increase the overall performance of the function. To that point, we probably shouldn't instantiate a new array `['0', 'false', 'nan']` each function call either. So to compromise, we can take advantage of precomputation while maintaining readability, I suggest we create a constant variable outside the `parseBool()` function body that runs once.
```ts
export const FALSE_TRUTHY_STRINGS = [0, false, NaN].map((v) => String(v).toLowerCase());

/** parseBool() */
```

How does this sound to you?

src/lib/common/utils.test.ts
EthanThatOneKid: Done <https://github.com/EthanThatOneKid/acmcsuf.com/commit/7b88b6be5fdff6bf021224e69fc0834b2adb490e>.

jaasonw: poggers

src/lib/common/utils.ts
jaasonw: I think we're talking nanoseconds of performance optimization at this point but, I believe that array would still get reinstantiated due to the stateless and ephemeral nature of serverless functions?

jaasonw (approved):

src/lib/common/utils.ts
EthanThatOneKid: This is true, but <https://github.com/EthanThatOneKid/acmcsuf.com/pull/455/commits/9e8decb4720a966d97dea67a611f34d0bc60d31b>.

src/lib/common/utils.ts
EthanThatOneKid: Resolved by <https://github.com/EthanThatOneKid/acmcsuf.com/pull/455/commits/9e8decb4720a966d97dea67a611f34d0bc60d31b>.
Edit: <https://github.com/EthanThatOneKid/acmcsuf.com/commit/4d40cb7e671a69de0a5811c1fc5df2512c4661d1>.

src/lib/common/utils.ts
jaasonw: 👍

src/lib/common/utils.ts
jaasonw: 👍

PR #466, opened 2022-05-12
Blog: Added filter by tags

### LINKED ISSUE #277: Blog: Create filter-by-tag feature
## Description
We're planning to have a filter by tag feature using Github GraphQL API or filter based on label with string search

## Resources
https://docs.github.com/en/graphql/overview/explorer

NLTN:
* Created TagField Component
	File Location: src/lib/components/utils/acm-tagfield.svelte
	Usage: Allows users to make selections and filter content.

* Created new file utils.ts to store common functions
	File Location: src/lib/common/utils.ts

* Added Filter by Tags feature for Blog page.
	Made some changes to these files:
	- In blog/index.svelte: Added TagField component.
	- In blog/_query.ts: Modified fetchNewsletters.
	- In blog/index.json.ts: Modified get function.

This PR is intended to resolve #277


src/routes/blog/index.svelte
```diff
@@ -1,9 +1,25 @@
 <script lang="ts" context="module">
   import type { LoadInput, LoadOutput } from '@sveltejs/kit/types/internal';
+  import TagField from '$lib/components/utils/acm-tagfield.svelte';
```
EthanThatOneKid: Nit: Import the `TagField` component along with the rest of the component import statements in the 2nd `<script>`.

src/routes/blog/index.xml.ts
```diff
@@ -53,7 +53,8 @@ function makeRssFeed(posts: Newsletter[]): string {
 }
 
 export async function get(): Promise<RequestHandlerOutput> {
-  return new Response(JSON.stringify(makeRssFeed(await fetchNewsletters())), {
+  const posts: Newsletter[] = (await fetchNewsletters()).posts;
```
EthanThatOneKid: Nit: Fancy [_deconstructor syntax_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) simplification.
```suggestion
  const { posts } = await fetchNewsletters();
```

src/lib/components/utils/acm-tagfield.svelte
```diff
@@ -0,0 +1,129 @@
...
+  function createTagURL(tag: string) {
+    if (urlSearchParamKey) {
+      if (selectedTags.includes(tag)) {
+        return `?${urlSearchParamKey}=${selectedTags.filter((t) => t !== tag).join(',')}`;
+      } else {
+        return `?${urlSearchParamKey}=${[...selectedTags, tag].join(',')}`;
+      }
+    } else {
+      return '';
+    }
```
EthanThatOneKid: ## Suggestions for `createTagURL()`

### 1) Nit: To avoid bugs, let's make the condition in the if-statement more strict.

```ts
if (urlSearchParamKey === '') {
```

### 2) To keep from making too many indents, let's short circuit.

```ts
if (urlSearchParamKey === '') {
  return '';
}
```

### 3) Let's determine the `nextTags` to be selected before constructing the string.

```ts
const nextTags = selectedTags.includes(tag)
      ? selectedTags.filter((t) => t !== tag)
      : selectedTags.concat([tag]);
```

### 4) Let's take advantage of the `URLSearchParams` API.

```ts
const params = [[urlSearchParamKey, nextTags.join(',')]];
return '?' + new URLSearchParams(params).toString();
```

### 5) Now, let's put it all together.

```suggestion
    if (urlSearchParamKey === '') {
      return '';
    }
    const nextTags = selectedTags.includes(tag)
      ? selectedTags.filter((t) => t !== tag)
      : selectedTags.concat([tag]);
    const params = [[urlSearchParamKey, nextTags.join(',')]];
    return '?' + new URLSearchParams(params).toString();
```

src/routes/blog/_query.ts
```diff
@@ -103,6 +113,14 @@ export async function fetchNewsletters(): Promise<Newsletter[]> {
     body: JSON.stringify({ query: newslettersQuery }),
   });
 
-  const newsletters = formatNewsletters(await response.json());
-  return newsletters;
+  let newsletters = formatNewsletters(await response.json());
+  const tags = discernTags(newsletters);
+
+  if (options && options.labels.length > 0) {
```
EthanThatOneKid: Nit: Strictly assert that the `options` object is defined.
```suggestion
  if (options !== undefined && options.labels.length > 0) {
```

src/routes/blog/_query.ts
```diff
@@ -103,6 +113,14 @@ export async function fetchNewsletters(): Promise<Newsletter[]> {
     body: JSON.stringify({ query: newslettersQuery }),
   });
 
-  const newsletters = formatNewsletters(await response.json());
-  return newsletters;
+  let newsletters = formatNewsletters(await response.json());
```
EthanThatOneKid: Since we are expecting to return a property `posts`, let's name this variable `posts` for convenience.
```suggestion
  const posts = formatNewsletters(await response.json());
```

src/routes/blog/_query.ts
```diff
@@ -103,6 +113,14 @@ export async function fetchNewsletters(): Promise<Newsletter[]> {
...
+  const tags = discernTags(newsletters);
+
+  if (options && options.labels.length > 0) {
+    newsletters = newsletters.filter((post) =>
+      post.labels.some((item) => options.labels.includes(item))
+    );
+  }
+
+  return { tags: tags, posts: newsletters };
```
EthanThatOneKid: To be fancy, let's use this object construction shorthand syntax:
```suggestion
  return { tags, posts };
```

src/routes/blog/_query.ts
```diff
@@ -103,6 +113,14 @@ export async function fetchNewsletters(): Promise<Newsletter[]> {
     body: JSON.stringify({ query: newslettersQuery }),
   });
 
-  const newsletters = formatNewsletters(await response.json());
-  return newsletters;
+  let newsletters = formatNewsletters(await response.json());
+  const tags = discernTags(newsletters);
+
+  if (options && options.labels.length > 0) {
+    newsletters = newsletters.filter((post) =>
```
EthanThatOneKid: Do not forget to swap out the variable names in this line.
```suggestion
    posts = posts.filter((post) =>
```

src/lib/common/utils.ts
```diff
@@ -0,0 +1,11 @@
+interface WithLabels {
+  labels: string[];
+}
+
+export function discernTags(posts: WithLabels[]): string[] {
```
EthanThatOneKid: Let's create a couple test cases ([_example_](https://github.com/EthanThatOneKid/acmcsuf.com/blob/dcd14855e9e14b402b2c72d41f16a0b26d5f7bef/src/lib/common/utils.test.ts#L4)) in `/src/lib/common/utils.test.ts` for `discernTags()` to make sure we have the behavior we want.

src/routes/blog/index.svelte
```diff
@@ -1,9 +1,25 @@
...
 
-  export async function load({ fetch }: LoadInput): Promise<LoadOutput> {
-    const response = await fetch(`/blog.json`);
-    return { props: { posts: await response.json() } };
+  let tags = [];
+  let selectedTags: string[] = [];
```
EthanThatOneKid: These two variables should not be reachable from the `module` script. The `module` script is ran on the server and sometimes on the browser. Therefore, we want to remove the variable declarations here and move them down to the 2nd `<script>`.
```suggestion
```

src/routes/blog/index.svelte
```diff
@@ -1,9 +1,25 @@
...
+
+  export async function load(event: LoadInput): Promise<LoadOutput> {
```
EthanThatOneKid: Nit: Since we only use `event` to access `event.url`, let's deconstruct `url` out right away.
```suggestion
  export async function load({ url }: LoadInput): Promise<LoadOutput> {
```

src/routes/blog/index.svelte
```diff
@@ -1,9 +1,25 @@
...
+      selectedTags = event.url.searchParams.get('l').split(',');
+    } else {
+      selectedTags = [];
+    }
+
+    const response = await fetch(target.toString());
+    const blogOutput = await response.json();
+    tags = blogOutput.tags;
+
+    return { props: { posts: blogOutput.posts } };
```
EthanThatOneKid: We must pass information to the client by returning it as a returned property of the `load` function.
```suggestion
    const target = new URL('/blog.json', url);
    const rawLabels = url.searchParams.get('l');
    
    const selectedTags = [];
    if (rawLabels !== null && rawLabels.length > 0) {
      target.searchParams.set('l', rawLabels);
      selectedTags.push(...rawLabels.split(','));
    }

    const response = await fetch(target.toString());
    const { posts, tags } = await response.json();

    return { props: { posts, tags, selectedTags } };
```

src/routes/blog/index.svelte
```diff
@@ -12,10 +28,23 @@
   import Spacing from '$lib/components/sections/spacing.svelte';
 
   export let posts: Newsletter[] = [];
```
EthanThatOneKid, line 28: Expect there to be more properties passed down.
```suggestion
  export let posts: Newsletter[] = [];
  export let tags: string[] = [];
  export let selectedTags: string[] = [];
```

src/routes/blog/index.svelte
```diff
@@ -12,10 +28,23 @@
...
+
+  async function filterPosts(event: CustomEvent) {
+    const tags = event.detail;
+
+    window.history.replaceState({}, '', `${window.location.pathname}?l=${tags.join(',')}`);
```
EthanThatOneKid: Nit: `window` APIs are accessible without the `window` object.
```suggestion
    history.replaceState({}, '', location.pathname + '?l=' + tags.join(','));
```

src/routes/blog/index.svelte
```diff
@@ -12,10 +28,23 @@
...
+
+    const target = new URL('/blog.json', window.location.origin);
+    target.searchParams.set('l', tags.join(','));
+
+    const response = await fetch(target.toString());
+    const blogOutput = await response.json();
+    posts = blogOutput.posts;
```
EthanThatOneKid: What if this request fails?

EthanThatOneKid (changes requested):
This PR represents a very complicated and well thought out feature. I left several suggestions regarding Svelte/TypeScript techniques. 

Do not hesitate to ask for clarification for any of the comments I left. Please reply to each comment with something like "Done." so that I can resolve the individual comment if it is ready to be resolved.

src/routes/blog/_query.ts
jaasonw: Alternatively, just `if (options?.labels.length > 0)`

jaasonw: More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

> The ?. operator is like the . chaining operator, except that instead of causing an error if a reference is [nullish](https://developer.mozilla.org/en-US/docs/Glossary/Nullish) ([null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null) or [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined)), the expression short-circuits with a return value of undefined. When used with function calls, it returns undefined if the given function does not exist.

src/routes/blog/_query.ts
jaasonw: Also it's good practice to use `const` whenever u know for a fact the variable will never be reassigned

src/routes/blog/index.json.ts
```diff
@@ -1,8 +1,14 @@
...
-import { fetchNewsletters } from './_query';
+import type { RequestHandlerOutput, RequestEvent } from '@sveltejs/kit/types/internal';
+import { fetchNewsletters, NewsletterFetchOptions } from './_query';
 
-export async function get(): Promise<RequestHandlerOutput> {
-  return new Response(JSON.stringify(await fetchNewsletters()), {
+export async function get(event: RequestEvent): Promise<RequestHandlerOutput> {
+  const fetchOptions: NewsletterFetchOptions = { labels: [] };
+
+  if (event.url.searchParams.has('l') === true && event.url.searchParams.get('l').length > 0) {
```
jaasonw: I'm of the opinion that checking === true is redundant, just do `if (event.url.searchParams.has('l') && event.url.searchParams.get('l').length > 0)`

conveys the same information whilst being more concise

src/routes/blog/_query.ts
NLTN: Thank you @EthanThatOneKid for pointing out. 
I like your solution @jaasonw. Tested and it worked as expected. Thanks

src/routes/blog/_query.ts
NLTN: I used `let` because it will be reassigned at line 120.

jaasonw: oop- yea i just noticed LOL I didnt read that far down yet

src/routes/blog/index.svelte
NLTN: I put those two variables in `module` script on purpose. Their values will be set in the `load()` function.

src/routes/blog/index.svelte
NLTN: Could you please explain more? Why do we need to `export` the variables `tag` & `selectedTags`?

NLTN: Yup, I will fix it. Now I feel like I begin to understand Svelte.

src/routes/blog/index.svelte
NLTN: Yup, I will fix it. Now I feel like I begin to understand Svelte.

src/routes/blog/_query.ts
EthanThatOneKid: Good catch! Sounds good.

src/routes/blog/index.svelte
NLTN: Good catch. Thanks.

src/lib/common/utils.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  expect(utils.discernTags([{ labels: ['algo'] }])).toEqual(['algo']);
+  expect(
+    utils.discernTags([
+      { labels: [] },
+      { labels: ['george washington'] },
+      { labels: ['victor', 'delta', 'delta', 'delta'] },
+      { labels: ['kilo', 'victor'] },
+      { labels: ['delta', 'alpha', 'bravo'] },
+    ])
+  ).toEqual(['alpha', 'bravo', 'delta', 'george washington', 'kilo', 'victor']);
```
EthanThatOneKid: 1. The name of the unit test should describe the behavior we are checking for.
2. We should break up the unit tests to make them more readable.
3. We should use simpler sample data to test with (which makes it easier to understand at a glance).

Below is a unit test that tests the behavior expected on line 7 (testing the case where multiple labels are discerned in alphabetical order). The unit test below can be used as an example to make separate unit tests for the expectations on lines 5 (testing the case where the argument has no labels) and 6 (testing the case where 1 label is discerned).

```suggestion
test('discerns unique labels from list in alphabetical order', () => {
  const input = [
    { labels: [] },
    { labels: ['abc def'] },
    { labels: ['d', 'c', 'c', 'c'] },
    { labels: ['e', 'd'] },
    { labels: ['c', 'a', 'b'] },
  ];
  const expected = ['a', 'abc def', 'b', 'c', 'd', 'e'];
  const output = utils.discernTags(input);
  expect(output).toEqual(expected);
```

**Note**: Make sure you have 2 more unit test definitions representing the expectations on lines 5 and 6.

src/lib/components/utils/acm-tagfield.svelte
```diff
@@ -0,0 +1,128 @@
...
+    return '?' + params.toString();
+  }
+</script>
+
+<div class="tag-box">
+  <div class="tag-title" class:hidden={hasSelectedTags}>
+    <slot name="title" />
+  </div>
+  <a href="?" class="tag-clear-button" class:hidden={!hasSelectedTags} on:click={deselectAll}>
+    <slot name="resetButton" />
```
EthanThatOneKid: For slot names, I do not believe we currently have decided on a naming convention. I personally think `name="reset-button"` with a `-` (hyphen; kebab-case) looks looks nicer in HTML markup. 
```suggestion
    <slot name="reset-button" />
```

EthanThatOneKid: There are currently [no occurrences](https://cs.github.com/EthanThatOneKid/acmcsuf.com?q=%3Cslot) of slots with multi-part names in this repository (this would be the first). So to be clear, we will use kebab-case for the slot names when we encounter a multi-part slot name.

src/routes/blog/index.svelte
NLTN: I fixed it.

src/routes/blog/index.svelte
NLTN: I fixed it.

src/routes/blog/index.svelte
NLTN: I fixed it.

src/routes/blog/index.svelte
NLTN: I fixed it.

src/lib/common/utils.test.ts
NLTN: The unit test has been updated.

src/lib/components/utils/acm-tagfield.svelte
NLTN: Absolutely agree. It looks nicer.

src/routes/blog/index.json.ts
```diff
@@ -1,8 +1,14 @@
...
+    fetchOptions.labels = event.url.searchParams.get('l').split(',');
+  }
```
EthanThatOneKid: Nit: Simplification of conditional statement.
```suggestion
  const fetchOptions: NewsletterFetchOptions = { labels: [] };
  const rawLabels = event.url.searchParams.get('l');

  if (rawLabels?.length > 0) {
    fetchOptions.labels = rawLabels.split(',');
  }
```

src/routes/blog/index.svelte
```diff
@@ -1,21 +1,48 @@
...
 
-  export async function load({ fetch }: LoadInput): Promise<LoadOutput> {
-    const response = await fetch(`/blog.json`);
-    return { props: { posts: await response.json() } };
+  export async function load({ url }: LoadInput): Promise<LoadOutput> {
+    const target = new URL('/blog.json', url);
+    const rawLabels = url.searchParams.get('l');
+    const selectedLabels = [];
+
+    if (rawLabels !== null && rawLabels.length > 0) {
```
EthanThatOneKid: Nit: Simplification of conditional statement. 
```suggestion
    if (rawLabels?.length > 0) {
```

src/routes/blog/index.xml.ts
```diff
@@ -53,7 +53,8 @@ function makeRssFeed(posts: Newsletter[]): string {
 }
 
 export async function get(): Promise<RequestHandlerOutput> {
-  return new Response(JSON.stringify(makeRssFeed(await fetchNewsletters())), {
+  const { posts } = await fetchNewsletters();
+  return new Response(JSON.stringify(makeRssFeed(posts)), {
```
EthanThatOneKid: Is `JSON.stringify` necessary on line 57?

src/lib/components/utils/acm-labelfield.svelte
```diff
@@ -0,0 +1,128 @@
...
+    }
+    @media screen and (max-width: 900px) {
+      .title {
+        display: none;
+      }
+      .reset-button {
+        display: none;
+      }
+    }
+  }
```
EthanThatOneKid: Nit: SCSS style definition spacing to ease the eyes (you may [hopefully] just directly commit this suggested change).
```suggestion
  .label-box {
    display: flex;
    align-items: center;
    justify-content: center;
    margin-bottom: 0.2em;

    .title {
      font-size: 1em;
      font-weight: bold;
      margin-right: 1em;
      cursor: default;
    }

    .hidden {
      display: none;
    }
    
    .reset-button {
      margin-right: 1em;
      font-size: small;
      text-decoration: underline;
      cursor: pointer;
    }
    
    .label-list {
      display: flex;
      flex-wrap: wrap;
      justify-content: center;
      margin-bottom: 0.2em;
      
      .label {
        margin-bottom: 0.2em;
        margin-right: var(--size-sm);
        padding: 0.25em 0.8em 0.25em 0.8em;
        background-color: #eeeeee;
        color: #212121;
        border-radius: var(--size-sm);
        border: 2px solid #e0e0e0;
        cursor: pointer;
        transition: 0.25s ease-in-out;
        text-decoration: none;
        
        &:hover {
          border-color: #b0bec5;
        }
        
        &.selected {
          transition: 1s ease-in-out;
          background-color: #81d4fa;
          border-color: #4fc3f7;
          
          &:before {
            content: '✓ ';
          }
          
          &:hover {
            border-color: #2196f3;
          }
        }
      }
    }
    
    @media screen and (max-width: 900px) {
      .title {
        display: none;
      }
      
      .reset-button {
        display: none;
      }
    }
  }
```

EthanThatOneKid (commented):
The code in this PR overall looks amazing! I left a couple new nitpicks related to @jaasonw's <https://github.com/EthanThatOneKid/acmcsuf.com/pull/466#discussion_r870944856> and one simply making a suggestion to space out your SCSS blocks. I cannot wait to see the search bar in production :) Getting this feature to work with JavaScript disabled was no easy task!

src/routes/blog/index.xml.ts
NLTN: Removed unnecessary code on line 57.

jaasonw:
I'll work on merging this in with my changes some time this week

EthanThatOneKid (approved):

jaasonw (changes requested):
It does not appear to be workin

image

src/routes/blog/index.json.ts
NLTN: > I'm of the opinion that checking === true is redundant, just do `if (event.url.searchParams.has('l') && event.url.searchParams.get('l').length > 0)`
> 
> conveys the same information whilst being more concise

You’re right. I fixed it couple weeks ago.

NLTN:
> It does not appear to be workin
> 
> image

I don’t know what happened. Maybe you’re in DEBUG mode or maybe there are something wrong in “.env” file.
Everything works perfectly on my PC.

jaasonw:
@NLTN Oh that's not my local environment it's the vercel preview linked in https://github.com/EthanThatOneKid/acmcsuf.com/pull/466#issuecomment-1124441436, if you want we can try to find some time to work on this maybe this week or a week from now (the week of 7/10)? If not we can push this back to when the semester starts

src/routes/blog/index.json.ts
```diff
@@ -1,15 +1,28 @@
...
+    import.meta.env.VITE_GH_ACCESS_TOKEN !== undefined &&
+    import.meta.env.VITE_GH_DISCUSSION_CATEGORY_ID !== undefined;
+
+  return new Response(
+    DEBUG || !isSatisfied ? SAMPLE_POSTS : JSON.stringify(await fetchNewsletters(fetchOptions)),
+    {
+      status: 200,
+      headers: { 'Content-Type': 'application/json' },
+    }
+  );
```
NLTN: @jaasonw Could you please check the Environment Variables in Vercel? I think we need to have these variables: VITE_GH_ACCESS_TOKEN, VITE_GH_DISCUSSION_CATEGORY_ID, and DEBUG. I don’t have a permission to check it.

jaasonw: Pinging @EthanThatOneKid for this

src/routes/blog/index.svelte
```diff
@@ -1,24 +1,51 @@
...
+    const rawLabels = url.searchParams.get('l');
+    const selectedLabels = [];
+
+    if (rawLabels?.length > 0) {
+      target.searchParams.set('l', rawLabels);
+      selectedLabels.push(...rawLabels.split(','));
+    }
+
+    const response = await fetch(target.toString());
+    const { posts, labels } = await response.json();
```
jaasonw: @NLTN Looks like console.logging the result of these variables returns 
```{ posts: undefined, labels: undefined }```
And that response only contains a list of blog posts with no labels. Did you maybe forget to commit a file?

jaasonw (changes requested):

src/routes/blog/index.svelte
NLTN: @jaasonw I double checked, and I believe that I committed all the necessary files. BTW, everything should work smoothy no matter what a list of blog posts has labels or not.

EthanThatOneKid (approved):

jaasonw (approved):
👍

PR #475, opened 2022-05-19
Events: Set up simple ICAL parser integration testing

EthanThatOneKid:
- Generated 10 events from the existing ` ./src/routes/events/_testdata/events.ics` into ` ./src/routes/events/_testdata/events.ts`.

https://github.com/EthanThatOneKid/acmcsuf.com/blob/5745f3efd890b9884ca9c4b7b0a145d5e941e31c/src/lib/ical/parse.test.ts#L15-L20

- Created `parse.test.ts` and within it, a test which verifies that our parser parses the ICAL data deterministically and expectedly.


k4rni (approved):
:)

jaasonw (commented):
poggers

PR #479, opened 2022-05-22
Events: Display TBD if location is empty or null

### LINKED ISSUE #416: I feel like default location should no longer be "Hosted on Discord"
### What happened?

image

charliettaylor:
### event-item
- checks if location is null or empty, displaying TBD if it is null, empty, or white space

hopefully small and easy fix :^)

Resolves #416


k4rni:
What an absolute king taking into consideration all edge cases and treating it like a leetcode problem

src/lib/components/events/event-item.svelte
```diff
@@ -49,6 +49,8 @@
         <p class="event-location">
           {info.location === 'Discord' || info.location === 'Zoom'
             ? `Hosted on ${info.location}`
+            : info.location == null || info.location.match(/^ *$/) !== null
+            ? 'TBD'
```
jaasonw: I think replacing this entire ternary with simply `${info.location?.trim() || "TBD"}` is more concise and covers all the edge cases as trim should reduce any spaces to empty strings that evaluate to falsey and a null should short circuit to undefined

jaasonw (changes requested):

src/lib/components/events/event-item.svelte
charliettaylor: Do we want Discord and Zoom events to still be displayed as "Hosted on X" tho?

src/lib/ical/parse.test.ts
```diff
@@ -15,5 +15,19 @@ test('ICAL parser integration test', () => {
...
+        prop !== 'location' &&
+        prop !== 'calendarLinks'
+      ) {
+        expect(element[prop]).toEqual(EVENT_DATA[i][prop]);
+      }
+    }
+    i += 1;
+  });
+
+  //expect(JSON.stringify(events)).toBe(JSON.stringify(EVENT_DATA));
```
charliettaylor: This is pretty weird but we have to skip location since it is changed within utils, also for some reason calendarLinks was being strange but that might just be because I don't know Jest too well

src/lib/components/events/event-item.svelte
jaasonw: I guess we could keep that part? But also why not just put the "Hosted on" in the actual calendar?

src/lib/ical/utils.ts
```diff
@@ -307,10 +307,18 @@ export function makeAcmEvent(
...
+  let location = locations.location;
+  const meetingLink = locations.meetingLink;
+
+  const hosted = ['Discord', 'Zoom'];
+
+  location = location?.trim()
+    ? hosted.includes(location.trim())
+      ? `Hosted on ${location}`
+      : location
+    : 'TBD';
```
jaasonw: Honestly if we're gonna keep the Hosted on, I think using an if statement would better help readability in this case.
```
if (hosted.includes(location.trim()))
  location = `Hosted on ${location}`
```

Also no need for the TBD or trim here because shouldn't event-item already trim it and/or make it a TBD? In which case why not move this logic back to event-item?

jaasonw (dismissed):

src/lib/ical/utils.ts
charliettaylor: We should keep the logic in here so we can have testing done to this and have expected results, I'll make a quick change to make it more readable though

src/lib/components/events/event-item.svelte
```diff
@@ -47,9 +47,7 @@
         </h2>
 
         <p class="event-location">
-          {info.location === 'Discord' || info.location === 'Zoom'
-            ? `Hosted on ${info.location}`
-            : info.location}
+          {info.location?.trim() || 'TBD'}
```
jaasonw: In which case wouldnt this be redundant?

src/lib/ical/utils.ts
EthanThatOneKid: I feel like our formatting operation, prepending “Hosted on”, could be handled in the Svelte file. Off-topic, but Angular has [pipes](https://angular.io/guide/pipes) for the technique I’m describing. 

The alternative is to add a new property to the `AcmEvent` interface, other than `AcmEvent.location`, which stores the formatted hosted-by string, something like `AcmEvent.hostedBy`.

src/lib/components/events/event-item.svelte
charliettaylor: Updated per conversation with Ethan, UI will do formatting and parse will handle null checks

jaasonw: i see

src/lib/components/events/event-item.svelte
```diff
@@ -21,6 +21,13 @@
       .catch(() => toast({ path, type: ToastType.Error, content: errorMessage }));
   }
 
+  function formatLocation(location?: string | null, hosted = ['Discord', 'Zoom']): string {
+    // '', null, and undefined are all TBD
+    location = location?.trim() ?? '';
+    if (!location) return 'TBD';
```
EthanThatOneKid: Nit: Since `location` always asserts a typeof `string` we can check specifically to see if the string is empty. 
```suggestion
    if (location === '') return 'TBD';
```

src/lib/ical/utils.ts
```diff
@@ -241,9 +241,13 @@ export function makeOutlookCalendarLink(
 
 export function parseLocation(
   rawLocation?: string,
-  defaultLocation = 'Discord',
+  defaultLocation = 'TBD',
   defaultLink = '/discord'
 ): { location: string; meetingLink: string } {
+  if (!rawLocation) {
+    rawLocation = rawLocation.trim();
+  }
```
EthanThatOneKid: Nit:
```suggestion
  rawLocation = rawLocation?.trim() ?? '';
```
Note: Now that `rawLocation` always asserts a typeof `string` we can remove the `?.` from the other conditionals in this function.

src/lib/ical/parse.test.ts
EthanThatOneKid: FYI, I originally used `JSON.stringify` to hash large lists of JavaScript objects and deterministically verify they matched. However, Vitest seems to give us access to a deep-equality method, `.toEqual()` (instead of `.toBe()`), which may work for this situation.

EthanThatOneKid (commented):
Looking cool so far @Peekoe 😎

src/lib/components/events/event-item.svelte
charliettaylor: ooo true, good catch

src/lib/ical/utils.ts
charliettaylor: Will make a commit with removing the other null operators

src/lib/ical/parse.test.ts
charliettaylor: I tried `.toEqual()` and it wasn't seeming to work, will try again tho during this next commit

src/lib/ical/parse.test.ts
```diff
@@ -15,5 +15,19 @@ test('ICAL parser integration test', () => {
   //   `export const events = ${JSON.stringify(events)};`
   // );
 
-  expect(JSON.stringify(events)).toBe(JSON.stringify(EVENT_DATA));
+  let i = 0;
+  events.forEach((element) => {
```
EthanThatOneKid: FYI ES6 array functions, like forEach, provide an index for you so you don’t have to manually increment it. 
```suggestion
  events.forEach((element, i) => {
```

charliettaylor: No longer need this for each but I shall remember this info next time I use it!

src/lib/components/events/event-item.svelte
```diff
@@ -21,6 +21,13 @@
       .catch(() => toast({ path, type: ToastType.Error, content: errorMessage }));
   }
 
+  function formatLocation(location?: string | null, hosted = ['Discord', 'Zoom']): string {
+    // '', null, and undefined are all TBD
```
jaasonw: can we get a unit test on this function for all 3 of these cases + the 2 base cases?

jaasonw (changes requested):

src/lib/components/events/event-item.svelte
charliettaylor: test cases for all TBD and online locations added!

src/lib/components/events/event-item.test.ts
```diff
@@ -0,0 +1,14 @@
+import { expect, test } from 'vitest';
+import { formatLocation } from './utils';
+
+test('can format bad locations', () => {
+  expect(formatLocation(null)).toBe('TBD');
+  expect(formatLocation('')).toBe('TBD');
+  expect(formatLocation(undefined)).toBe('TBD');
+  expect(formatLocation("   Your mom's house  ")).toBe("Your mom's house");
```
jaasonw, line 8: How about "can format good locations"? and should this fall under that condition?

jaasonw (changes requested):

src/lib/components/events/event-item.test.ts
charliettaylor: Technically this isn't a good location, since it has white space on either end. Any data that has been parsed correctly should already be trimmed

jaasonw: this is true, howabout a good location test case tho?

charliettaylor: New tests added

jaasonw: lgtm : )

jaasonw (approved):

PR #492, opened 2022-06-04
Restyle tags and rearrange author and read time

jaasonw:
image

Resolves
image


anhduy1202:
### Hmm, the hashtag position above author seems unusual to me, what do yall think about this design?

image

anhduy1202 (approved):
lgtm

PR #494, opened 2022-06-06
Added code block support to the blog

### LINKED ISSUE #487: add code block support to blog
### What's your idea?

also include syntax highlighting pls thanks

jaasonw:
Currently looks like this
image
image

Would like feedback on
a. What color we should use for the background (they're copy pasted from github along with the syntax highlighting)
b. How to make syntax-highlight.css less repetitive as scss does not seem to work??

I also updated the test cases and test data to reflect these changes


k4rni:
cool.

k4rni (approved):
try hard

PR #496, opened 2022-06-08
fade bottom edge of blog preview

jaasonw:
Just a 1 liner css property to fade the bottom edge of the blog preview as per @karnikaavelumani suggestion
image

Looks like this
image

### DIFF
--- a/src/routes/blog/index.svelte
+++ b/src/routes/blog/index.svelte
@@ -152,6 +152,7 @@
             max-height: 100px;
             overflow: hidden;
             margin: 16px 0;
+            mask-image: linear-gradient(to bottom, black 50%, transparent 100%);
           }
         }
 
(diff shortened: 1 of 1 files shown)


EthanThatOneKid:
This one(ish)-liner is impressive; thanks :)

We attempted something similar to this [in our old, custom event carousel component](https://github.com/EthanThatOneKid/acmcsuf.com/blob/ab993e7c07c12a459a9fbac8f8e7035b5d2dec4e/src/components/events/event-carousel.svelte#L121-L139). I recall it being quite tricky to get it "right" at the time.

**Note**: Small PRs like this are awesome for multiple reasons. This PR highlights one reason, which is its ability to share small [working/up-to-date] nuggets of code that astute members of _ACM Chapter at CSUF_ can learn and use in their own endeavors. We support ourselves and our colleagues by writing small, direct PRs.

k4rni:
This is better than what it is currently so I approve. If you have any ideas to redesign the blog to make it look even nicer we can also take that into consideration. Rn tho it's good to me.

k4rni (approved):

PR #497, opened 2022-07-08
add code coverage test

jaasonw:
Code coverage is a measure of how much of the code is being tested
See: https://www.atlassian.com/continuous-delivery/software-testing/code-coverage

This PR adds the command `npm run test:coverage` to generate a code coverage report after running all unit tests. It achieves this using c8 as per the vitest docs
https://vitest.dev/guide/coverage.html

Currently our code coverage looks like this
image
And we should definitely look to improve this metric going forward

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 /package
 .vercel_build_output
 .env
+/coverage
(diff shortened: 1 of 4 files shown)


janiljohn (approved):
looks poggers

PR #517, opened 2022-08-29
deprecate nodebuds

### LINKED ISSUE #516: deprecate node buds button
i guess we arent doing this anymore?

jaasonw:
only removing from navbar but keeping the page in the event we ever want to bring it back

image

image


k4rni (approved):
We can go ahead and add a "club collaboration" page in the near future.

PR #526, opened 2022-09-02
Add page for CS Survival Guide event

### LINKED ISSUE #520: Add CS Survival Guide page
### What's your idea?

We should have a CS Survival Guide page. It shall be a completely separate page from `/events`, which could probably be `/cssg`.

The page will be hidden, as in no hyperlinks will reach it by default. However, we can program it so that 1-2 weeks prior to the event start date, the website will either reveal a banner advertising the page or add a section into the main page advertising it. I think the banner is likely better.

To be discussed: what to be included in the cssg page. We can use @AlanCortez1337's flyer designs for this.

EthanThatOneKid:
This page will be on <https://acmcsuf.com/cssg/>.

Resolves #520.


jaasonw (approved):
👍

PR #532, opened 2022-09-07
Event transitions 🎉

JustinStitt:
### Demo
show1

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,2757 @@
+lockfileVersion: 5.4
+
+specifiers:
+  '@js-temporal/polyfill': ^0.4.1
+  '@playwright/test': ^1.27.1
+  '@size-limit/file': ^8.1.0
+  '@sveltejs/adapter-auto': next
+  '@sveltejs/kit': next
+  '@types/html-to-text': ^8.1.1
+  '@types/qrcode': ^1.5.0
+  '@types/rss': ^0.0.29
+  '@typescript-eslint/eslint-plugin': ^5.27.0
+  '@typescript-eslint/parser': ^5.27.0
+  '@vitest/coverage-c8': ^0.24.0
+  discord.js: ^13.6.0
+  dotenv: ^16.0.2
+  eslint: ^8.16.0
+  eslint-config-prettier: ^8.3.0
+  eslint-plugin-svelte3: ^4.0.0
+  html-to-text: ^8.2.1
+  prettier: ^2.6.2
+  prettier-plugin-svelte: ^2.7.0
+  qrcode: ^1.5.1
+  rrule: ^2.6.9
+  rss: ^1.2.2
+  sass: ^1.54.9
+  size-limit: ^8.1.0
+  svelte: ^3.44.0
+  svelte-check: ^2.7.1
+  svelte-confetti: ^1.2.0
+  svelte-preprocess: ^4.10.6
+  tslib: ^2.3.1
+  typescript: ^4.7.4
+  vite: ^3.1.0
+  vitest: ^0.23.2
+  web-vitals: ^3.0.3
+
+dependencies:
+  '@js-temporal/polyfill': 0.4.3
+  html-to-text: 8.2.1
+  qrcode: 1.5.1
+  rrule: 2.7.1
+  rss: 1.2.2
+  svelte-confetti: 1.2.2
+  web-vitals: 3.1.0
+
+devDependencies:
+  '@playwright/test': 1.29.1
+  '@size-limit/file': 8.1.0_size-limit@8.1.0
+  '@sveltejs/adapter-auto': 1.0.0-next.91_l5ueyfihz3gpzzvvyo2ean5u3e
+  '@sveltejs/kit': 1.0.0-next.589_svelte@3.55.0+vite@3.2.5
+  '@types/html-to-text': 8.1.1
+  '@types/qrcode': 1.5.0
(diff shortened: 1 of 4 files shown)


src/lib/components/events/event-item.svelte
```diff
@@ -57,16 +64,18 @@
         href={info.meetingLink}
         role="button"
         target="_blank"
-        rel="noopener noreferrer">Join</a
+        rel="noopener noreferrer"
+        on:click={(/* janky hack */) => {
```
JustinStitt: Without this _hack_ if the user clicks the "Join" button on an event it will toggle the `shown` state causing an inverse behavior, I added this to retoggle `shown` since clicking the "Join" button counts as clicking on the Event itself.

src/lib/components/events/event-item.svelte
```diff
@@ -57,16 +64,18 @@
...
       >
     </summary>
 
-    <hr />
-
-    <p class="event-description">
-      {@html info.description}
-    </p>
-
+    {#if shown}
```
diamondburned: This breaks without JavaScript. We should keep a `<noscript>` tag outside of the `if` block like so:

```html
<noscript>
  <p class="event-description">
    {@html info.description}
  </p>
</noscript>
{#if shown}
  <!-- same p tag as above -->
{/if}
```

diamondburned (changes requested):

diamondburned:
As a side note, the Algo event's description escapes the `<p>` tag, probably because nested `<p>` tags are invalid. It should probably be in a `<div>` instead. Note the `HTML_TAG_START` and `HTML_TAG_END` comments.

cc @EthanThatOneKid 

image

diamondburned:
Also, can we squash these commits?

JustinStitt:
> Also, can we squash these commits?

I like have PR revision history. Perhaps maintainer can squash pre-merge?

JustinStitt:
Also, this [package](https://github.com/SaekiTominaga/customelements-details-animation) may be of interest. It does detail-summary animations in-house.

Adds a bit of css bloat but simplifies the DOM.

JustinStitt (commented):
I have the power to LGTM but won't :)

diamondburned:
> As a side note, the Algo event's description escapes the `<p>` tag, probably because nested `<p>` tags are invalid. It should probably be in a `<div>` instead. Note the `HTML_TAG_START` and `HTML_TAG_END` comments.

After looking into it, this doesn't happen with JS enabled. It might be the server-side rendering attempting to sanitize the HTML differently compared to the browser.

jaasonw:
am i missing something or it seem kinda janky to me?

JustinStitt:
> am i missing something or it seem kinda janky to me?

The behavior or my implementation?

jaasonw:
the behavior

jaasonw:
https://user-images.githubusercontent.com/8981287/189256563-8a2cc082-ccc6-40f3-acef-154185aa3c03.mp4
like it just kinda snaps down and when u minimize it kinda just disappears

JustinStitt:
> https://user-images.githubusercontent.com/8981287/189256563-8a2cc082-ccc6-40f3-acef-154185aa3c03.mp4 like it just kinda snaps down and when u minimize it kinda just disappears

hmm, idk. I just tried to be clever and use the built-in Svelte transitions. It does look weird to me as well.

EthanThatOneKid (approved):
I appreciate the changes in this PR because @JustinStitt was able to create an enhancement that gets added to the page progressively when JavaScript is enabled. This means that he took into account backwards-compatibility, allowing the component to properly function even when the expected JavaScript code is absent.

diamondburned (approved):

src/lib/components/events/event-item.svelte
EthanThatOneKid: This comment may be used to create a follow up issue in the future. Resolving to merge the PR.

jaasonw:
wait hold up, am i the only one that finds this initial jerk kinda weird

https://user-images.githubusercontent.com/8981287/194151540-0bbb4174-672d-4b2e-b814-e4b3cd7819b9.mov

JustinStitt:
I am not seeing an initial jerk. It's animation timing is probably a Cubic Bezier with a moderately high starting slope but that is just part of the animation.

This is preferred to a slow, unresponsive, and less punchy _ease-in_ timing imo.

diamondburned:
> I am not seeing an initial jerk.

Probably because the animation timing is very low. By doing this change:

```diff
--- a/src/routes/events/event.svelte
+++ b/src/routes/events/event.svelte
@@ -93,7 +94,7 @@
     </noscript>
 
     {#if shown}
-      <p class="event-description" transition:slide>
+      <p class="event-description" transition:slide={{ duration: 2000 }}>
         {@html info.description}
       </p>
     {/if}
```

You'll see that the action buttons inside each event are not actually transitioned properly:

[Screencast from 05-10-22 15:24:01.webm](https://user-images.githubusercontent.com/8463786/194175000-de8e9d17-8ce4-4b1f-a0de-781498344284.webm)

This probably explains the initial jerk.

jaasonw:
oh, good catch, yea these buttons needa be transitioned as well as the description
image

jaasonw:
https://user-images.githubusercontent.com/8981287/209519742-0600c6d0-8de0-4ee4-a158-aca82df6ff70.mp4

jaasonw:
it works but i had to do it on this older version of the website because no gcal api key

PR #533, opened 2022-09-09
add arish

jaasonw:
manually adding arish in to the ai board list because the automation is still broken, placeholder still being used because no headshot was submitted


EthanThatOneKid (approved):
Welcome, Arish!

PR #575, opened 2022-09-21
SvelteKit release candidate migration: service-worker 

EthanThatOneKid:
This feature allows users to install our website to get info about ACM at CSUF while offline. There are no migration instructions for migrating the service worker in <https://github.com/sveltejs/kit/discussions/5774>. I referred to this example: <https://github.com/Myrmod/SvelteKit-offline/blob/c8cd6874c16bc838732189b0c4476b88f1a00bc0/src/service-worker/index.ts>.

### DIFF
--- a/src/service-worker/install.ts
+++ b/src/service-worker/install.ts
@@ -1,0 +1,16 @@
+/// <reference lib="webworker" />
+import { CURRENT_CACHE_NAME } from './common';
+import { build } from '$service-worker';
+
+export function handleInstall(event: ExtendableEvent): void {
+  event.waitUntil(
+    // Cache known build output and static files
+    // @see <https://kit.svelte.dev/docs/modules#$service-worker-build>
+    caches.open(CURRENT_CACHE_NAME).then((cache) => {
+      // Open a cache and cache our files
+      cache.addAll(build);
+
+      return true;
+    })
+  );
+}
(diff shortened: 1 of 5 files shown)


src/service-worker/install.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference lib="webworker" />
+import { CURRENT_CACHE_NAME } from './common';
+import { build } from '$service-worker';
+
+export function handleInstall(event: ExtendableEvent): void {
+  event.waitUntil(
+    // Cache known build output and static files
+    // @see <https://kit.svelte.dev/docs/modules#$service-worker-build>
+    caches.open(CURRENT_CACHE_NAME).then((cache) => {
```
charliettaylor, line 9: Nit: Do we have to use `then` here or can we use `await`?

EthanThatOneKid: Good observation. Async function handlers are probably allowed to be passed to the event listeners, but we currently do not have strict style guidelines for using `await` over `then`. One advantage of `then`, for example, is its ability to chain error handling (via `catch` and `finally`) instead of wrapping multiple lines with `try`/`catch`.

charliettaylor (approved):
For am I now seeking the approval of man, or of God? (Galatians 1:10)

EthanThatOneKid:
Replaced by mega-PR #577.

PR #576, opened 2022-09-21
Add fallback fonts and set font-display to fallback

### LINKED ISSUE #574: Bug: No fallback fonts
### What happened?

For users with low-bandwidth/slow Internet, acmcsuf.com will display the ugly Serif default font instead of the expected Poppins font. The site should account for this and provide fallback fonts in the [global.css](https://github.com/EthanThatOneKid/acmcsuf.com/blob/13dbf81066de81b7aad52a087259edc3534d06e6/static/global.css#L45):

https://github.com/EthanThatOneKid/acmcsuf.com/blob/13dbf81066de81b7aad52a087259edc3534d06e6/static/global.css#L42-L46

Since Poppins is a Sans-Serif font, we can expect the fallback to look something like `"Poppins", "Helvetica", sans-serif;`.

mikelee:
I added Verdana because it more closely matches Poppins per https://meowni.ca/font-style-matcher/

I also thought it would be a good idea to set the font-display to fallback. If you would prefer a different behavior, check out https://web.dev/font-best-practices/#choose-an-appropriate-font-display-strategy

Fixes #574


jaasonw (approved):
i like it 👍

EthanThatOneKid (approved):
Thank you for sharing your thought process. This looks great!

PR #591, opened 2022-09-22
Solved Legacy Quiz Data Issue

AlanCortez1337:
- ❌Removed the color attribute from QuizData since it's unnecessary.
- ✔Created new data.ts file to store more info for the teams.


EthanThatOneKid (approved):
Thank you for following up with maintaining the `/quiz` page.

PR #595, opened 2022-09-28
Add prefetch to layout for events

### LINKED ISSUE #530: events page loads slow
currently the events page loads really slow. this is because the following steps have to occur

1. wait for the serverless function to cold start
2. wait for the calendar api to return a response
3. generate the html page with the events already loaded

this creates a bottleneck where the api call essentially blocks the page load until it completes

i suggest changing the process to either load the events asynchronously as to not block the page load

1. load the page without the events
2. call the calendar api after or during the page load
3. add the events to the page afterwards

charliettaylor:
Resolves #530


src/routes/+layout.svelte
```diff
@@ -17,6 +18,10 @@
       mediaList.addEventListener('change', changeTheme);
       return () => mediaList.removeEventListener('change', changeTheme);
     }
+    
+    // prefetch used here because of the load time of events query, see #530
```
EthanThatOneKid:
```suggestion
    // prefetch used here to improve the load time of events query, see #530
```

EthanThatOneKid (approved):

src/routes/+layout.svelte
charliettaylor: Technically not improving load time of query, but of the events page. So should say "refetch used here to improve the load time of the events page, see #530"

src/routes/+layout.svelte
```diff
@@ -17,6 +18,10 @@
       mediaList.addEventListener('change', changeTheme);
       return () => mediaList.removeEventListener('change', changeTheme);
     }
+
+    // refetch used here to improve the load time of the events page, see #530
```
EthanThatOneKid:
```suggestion
    // prefetch improves the load time of the events page, see #530
```

EthanThatOneKid (approved):

PR #596, opened 2022-09-28
Add member metadata and flip card

diamondburned:
Design 1: the whole member card flips to reveal the backside of the card.

Sibling: https://github.com/EthanThatOneKid/acmcsuf.com/pull/597

For preview, go to the [About page](https://acm-csuf-site-git-member-card-ethanthatonekid.vercel.app/about). Flip cards are available for @EthanThatOneKid, @AlanCortez1337 and @diamondburned.

TODO

Issue #538, opened 2022-09-10
Dedicate team data for `/quiz`


EthanThatOneKid:
### What happened?

The data that currently supplies the `/quiz` page uses a legacy constant variable that should no longer be edited. A new "teams"-branded constant variable would suit our needs. In the meantime, the changes made to the constants in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/528/files#diff-9c15f43b01f835ed6e77ac6a576fdbcd72cdc5a6e9a928d1b3d153a61a6605d8> should be moved to another variable in `/src/lib/quiz/`.

k4rni:
Resolved by #591.

PR #597, opened 2022-09-28
 Add member metadata and flip card (2nd variant)

diamondburned:
Design 2: the small member image flips to reveal the backside of itself.

Sibling: https://github.com/EthanThatOneKid/acmcsuf.com/pull/596

For preview, go to the [About page](https://acm-csuf-site-git-member-card2-ethanthatonekid.vercel.app/about). Flip cards are available for @EthanThatOneKid, @AlanCortez1337 and @diamondburned.

TODO


diamondburned:
Flipped social card

src/routes/about/officer-profile.svelte
```diff
@@ -10,40 +17,150 @@
...
-    .replace(/Marketing\s/, `<b class="acm-red">Marketing&nbsp;</b>`)
-    .replace(/Special Events\s/, `<b class="acm-yellow">Special Events&nbsp;</b>`)
-    .replace(/nodebuds\s/, `<span class="brand-header">Node Buds&nbsp;</span>`);
+  $: [titleName, titleClass] = positionClasses.find((e) => {
+    return officerPosition.startsWith(e[0]);
+  }) || [null, 'acm-dark'];
+
+  $: titleHTML = titleName
+    ? officerPosition.replace(titleName, `<b class="${titleClass}">${titleName}</b>`)
+    : officerPosition;
```
EthanThatOneKid: Nice refactor.

src/lib/components/toaster/toasts.ts
```diff
@@ -64,3 +64,25 @@ export function postponeDismissal(id: number, timeout: number): void {
...
+export type doToastOpts<T> = Omit<Omit<Toast, 'id'>, 'content'> & {
+  success: (_: T) => string;
+  failure: (err: unknown) => string;
+};
+
+export async function toastDo<T>(
+  f: () => T | Promise<T>,
+  opts: doToastOpts<T> = {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    success: (_: T) => 'Success!',
```
EthanThatOneKid:
```suggestion
    success: () => 'Success!',
```

src/lib/components/toaster/toasts.ts
```diff
@@ -64,3 +64,25 @@ export function postponeDismissal(id: number, timeout: number): void {
 export function dismissToast(id: number): void {
   toasts.update((all) => all.filter((t) => t.id !== id));
 }
+
+export type doToastOpts<T> = Omit<Omit<Toast, 'id'>, 'content'> & {
+  success: (_: T) => string;
+  failure: (err: unknown) => string;
+};
+
+export async function toastDo<T>(
```
EthanThatOneKid: Nit: Verb to emit toast is not specific enough. 
Consider: <https://github.com/crcn/programming-terms#mediator> emit methods.

src/lib/components/toaster/toasts.ts
```diff
@@ -64,3 +64,25 @@ export function postponeDismissal(id: number, timeout: number): void {
...
+  f: () => T | Promise<T>,
+  opts: doToastOpts<T> = {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    success: (_: T) => 'Success!',
+    failure: (err: unknown) => `An error occured while performing action: ${err}`,
+  }
+) {
+  try {
```
EthanThatOneKid: If await is called on this function (e.g. `await toastDo(/*...*/);`), then it will block on this try/catch block. I do not think this currently causes any problems, but it is probably not intentional to block the parent scope on the toast function (future contributors may see "async" and want to await it, but they really only should when they want to block on the toast function). A non-blocking approach would call for then/catch over try/catch and return without waiting for the resolution.

src/routes/about/officer-profile.svelte
```diff
@@ -10,40 +17,150 @@
...
+
+  /** @see <https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText> */
+  function copyDiscord() {
+    toastDo(() => navigator.clipboard.writeText(officerSocials.discord || ''), {
+      success: () => `Copied ${officerName}'s Discord tag to clipboard!`,
+      failure: () => 'Error occured while copy Discord tag to clipboard.',
+    });
+  }
+
+  const positionClasses = Object.entries({
```
EthanThatOneKid: Nit: Provide a comment to clarify how this variable is used. 

Note: The value of positionClasses should be derived from a future common module containing predictable information about each team (related: #538).

EthanThatOneKid (commented):
This is a neat way of organizing board members socials on our about page. I left some initial comments.

src/lib/components/toaster/toasts.ts
diamondburned: Do is about doing the callback, as opposed to just `toast` (which is not even a verb). I don't think any of the verbs listed here are suitable.

src/lib/components/toaster/toasts.ts
diamondburned: This doesn't matter; the user does `await toastDo` if they wish to wait for the given callback. `toast` itself is not async, so this has virtually no impact.

src/lib/components/toaster/toasts.ts
EthanThatOneKid: Even then, we tend to place the verb before the subject.
 
Note: `toast` (and many nouns) may technically be used as a verb, though it is, admittedly, not so obvious depending on the context. I value code being easy on the brains of any programmer to understand at first glance (without further/unneeded/avoidable contextualization), hence the seemingly inane discussion on function naming.

diamondburned: > Even then, we tend to place the verb before the subject.

Because `doToast`, which is the common way to word functions, describes a completely different (but a lot more common) functionality. `toastDo` is not.

A longer name for this function would be `toastDoFunc` or `toastDoFn`.



> Note: `toast` (and many nouns) may technically be used as a verb, though it is, admittedly, not so obvious depending on the context.

I'm not aware of `toast` (verb) being used as a verb for spawning a toast notification anywhere. An actual, more common verb would be `popup`.

diamondburned: Fixed by making `toast` possibly accept a function.

EthanThatOneKid (approved):

src/routes/(site)/about/officer-profile.svelte
```diff
@@ -9,41 +17,168 @@
   export let info: Officer;
   export let placeholderPicture = 'placeholder.webp';
   export let dev = false;
```
EthanThatOneKid, line 19: Note: Variable used to determine whether the dev server is enabled can be imported like so:
```ts
import { dev } from '$app/environment';
```
See: <https://kit.svelte.dev/docs/modules#$app-environment-dev>.

diamondburned: I'm actually not sure why this was here in the first place. I didn't touch it in any meaningful way.

EthanThatOneKid: > I'm actually not sure why this was here in the first place. I didn't touch it in any meaningful way.

All good. I believe it is being followed up in #612.

Issue #557, opened 2022-09-14
Save screenshots as side-effect in Playwright tests


EthanThatOneKid:
### What's your idea?

To help keep track of all visual changes made to the website from now on, I propose that we use Playwright to take screenshots of each page configured to do so. In this proposal, I am considering taking the screenshots as a side-effect of the Playwright tests, since they do the job of spinning up each page.

Related: #336, #120

EthanThatOneKid:
Related: <https://playwright.dev/docs/test-configuration#automatic-screenshots>

PR #601, opened 2022-10-07
Use full names for blog post authors

### LINKED ISSUE #488: add full name to blog posts instead of github username
### What's your idea?

image
Currently the blog posts only return the display name but I'd like it to return the full name instead and maybe the bio?
image

charliettaylor:
Resolves #488

I removed officer since it doesn't work for anyone currently

### DIFF
--- a/src/lib/server/blog/posts.ts
+++ b/src/lib/server/blog/posts.ts
@@ -69,39 +69,41 @@ async function cacheBlogPosts(output: any): Promise<BlogPost[]> {
    * based on the GraphQL {@link gql}.
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const posts = await Promise.all(discussions.map(async (discussion: any): Promise<BlogPost> => {
-    const {
-      title,
-      author,
-      createdAt,
-      lastEditedAt: lastEdited,
-      number: id,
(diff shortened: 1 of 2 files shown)


EthanThatOneKid:
> I removed officer since it doesn't work for anyone currently

I think that the officer images regression occurred after a recent PR (maybe #597). 

Side note: Screenshot testing would allow us to catch unintended changes in other pages earlier. Related: #557.

src/lib/server/blog/posts.ts
```diff
@@ -61,56 +59,90 @@ export async function fetchBlogPosts(options?: BlogFetchOptions): Promise<BlogOu
...
+    createdAt,
+    lastEdited,
+    labels,
+    author: { url: authorUrl, displayname, picture },
+  };
+
+  return post;
+}
+
+// function getOfficerByGhUsername(ghUsername: string): Officer | null {
```
jaasonw, line 126: seem to work now for some reason, but not exactly, maybe look into fixing this or just commit to removing it entirely?

image

jaasonw (changes requested):

charliettaylor:
lol the officer data already works, we just don't have it. non board members will not have full names displayed to avoid API calls

PR #577, opened 2022-09-21
Migrate to SvelteKit release candidate

EthanThatOneKid:
Migration guide: <https://github.com/sveltejs/kit/discussions/5774>

### What this PR accomplishes

- Recreation of the website based on `npm create svelte@latest acmcsuf.com` (ran on 9/10/22)
- Moved [TypeScript] library/[Svelte] component files from `/src/lib/` to `/routes/` directory. Introduction of `/lib/public/legacy/` for extraneous files.
- Copied over docs, assets, and static content into `/` and `/static/`
- Renders the `/` index page
- Migrates `/src/service-worker/` (<https://kit.svelte.dev/docs/service-workers>)
- Migrates `/src/params/` (<https://kit.svelte.dev/docs/advanced-routing#matching>)
- Migrates error page, `/src/routes/+error.svelte` (<https://kit.svelte.dev/docs/routing#error>)
- Migrates Svelte hooks (<https://kit.svelte.dev/docs/hooks#server-hooks>)
- Disobeys the old file architecture

### What this PR consists of

This mega-PR consists of the following smaller PRs:

1. #540 Reviewed by @karnikaavelumani 
1. #541 Reviewed by @karnikaavelumani 
1. #554 Reviewed by @diamondburned 
1. #572 Reviewed by @karnikaavelumani 
1. #555 Reviewed by @erwinmedina 
1. #556 Reviewed by @Angus-1 
1. #558 Reviewed by @AlanCortez1337 
1. #559 Reviewed by @diamondburned 
1. #561 Reviewed by @Angus-1 
1. #568 Reviewed by @jaasonw 
1. #571 Reviewed by @jaasonw 
1. #573 Reviewed by @jaasonw 
1. #575 Reviewed by @Peekoe

Resolves #502.

PR #603, opened 2022-10-07
Moved all existing routes to route group `(site)`

EthanThatOneKid:
Reorganizes all of the contents of `/src/routes/` into a new _group_ under `/src/routes/(site)/`. This allows future pages to be created without rendering the default site layout, by using a reset layout.

Excerpt from the new `src/routes/+layout.svelte`:
```
// This +layout.svelte is a blank layout. It is used by the
// pages that don't need a layout. It is also used as the base
// layout of all pages.
```

More info: <https://kit.svelte.dev/docs/advanced-routing#advanced-layouts>.
Note: Route grouping is now available after migrating to the SvelteKit release candidate (related: #577).


k4rni (approved):
Very convenient

PR #607, opened 2022-10-08
Shortlinks: Add smart shortlink resolution

### LINKED ISSUE #592: Shortlinks: Allow subdirectories and appending
### What's your idea?

Shortlinks on `acmcsuf.com` currently match and redirect by exact ID.

By allowing subdirectories, we can make more organized, predictable shortlinks. With subdirectories, we could update some of our shortlinks (for example, `acmcsuf.com/design-template` to `acmcsuf.com/design/template`).

With appending, visitors may append a `/` to any known shortlink followed by any path. This would allow for us to utilize shortlinks like `acmcsuf.com/github` in convenient ways, such as `acmcsuf.com/github/issues/`, `acmcsuf.com/github/blob/main/ARCHITECTURE.md`, `acmcsuf.com/github/raw/main/package.json`, and more.

EthanThatOneKid:
- Introduced interface `Link<ID extends string>` in './types.ts'. This interface is implemented as a linked list.
- Introduced `validate` function in './validate.ts' and tests in './validate.ts'. This function validates shortlinks and constructs the `Link<ID extends string>` linked list if valid.
- Introduced `resolve` function in './resolve.ts' and tests in './resolve.test.ts'. Resolves shortlinks and computes the absolute destination URL.

Intended to resolve #592.


src/lib/server/links/utils.ts
```diff
@@ -1,15 +1,70 @@
...
+  let hashStr = '';
+  if (hashBeginIdx !== -1) {
+    hashStr = url.slice(hashBeginIdx + 1);
+    url = url.slice(0, hashBeginIdx);
+  }
+
+  const queryBeginIdx = url.indexOf('?');
+  let queryStr = '';
+  if (queryBeginIdx !== -1) {
+    queryStr = queryBeginIdx === -1 ? '' : url.slice(queryBeginIdx + 1);
```
jaasonw: ternary seem redundant, doesnt if statement on line 14 already guarantee queryBeginIdx not -1?

src/lib/server/links/utils.ts
```diff
@@ -1,15 +1,70 @@
...
+ * @param subCollection The collection of links to search.
+ * @returns The link object.
+ */
+export function parseLinkId<ID extends string>(
+  url: string,
+  subCollection: Record<ID, string>,
+  separator = '/'
+): Link<ID> | undefined {
+  const link = parseLink(url);
+  if (!link) return;
```
jaasonw: can we unit test this condition?

src/lib/server/links/utils.ts
```diff
@@ -1,15 +1,70 @@
...
+
+  const id = findLinkId(link.relativePathname.split(separator), subCollection, separator);
+  if (id === undefined) return;
```
jaasonw: also this one?

src/lib/server/links/utils.ts
```diff
@@ -1,15 +1,70 @@
...
+): ID | undefined {
+  let id = segments.join(separator) as ID;
+  while (id.length > 0) {
+    if (subCollection[id]) return id;
+    id = id.slice(0, id.lastIndexOf(separator)) as ID;
+  }
+}
+
+/**
+ * Converts a URL to a link object.
```
jaasonw: function could be document better i think, cannot tell what this does and look at implementation for link object not helpful

src/lib/server/links/utils.ts
```diff
@@ -1,15 +1,70 @@
...
+  const relativePathnameStr = url
     .replace(/^\/+|\/+$/g, '')
     .replace(/[^a-zA-Z0-9-/]/g, '')
     .toLowerCase();
 
-  if (!(cleanedLinkName in LINKS)) return;
+  return { relativePathname: relativePathnameStr, query: queryStr, hash: hashStr };
+}
+
+export function findLinkId<ID extends string>(
```
jaasonw: document this function

jaasonw (changes requested):

src/lib/server/links/utils.ts
EthanThatOneKid: I temporarily intended to use a ternary for this operation, but decided on an if-statement instead. I seem to have forgor to fix the ternary; good catch. Will fix in the morning.

src/lib/server/links/utils.ts
jaasonw: maybe add context on what it's for somewhere

src/lib/server/links/utils.ts
EthanThatOneKid: Resolved in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/607/commits/b7d7b68bbfe8a9ba96f567b21907f756c5cf101f>.

src/lib/server/links/utils.ts
EthanThatOneKid: Progress made in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/607/commits/b7d7b68bbfe8a9ba96f567b21907f756c5cf101f>.

src/lib/server/links/utils.ts
EthanThatOneKid: Resolved in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/607/commits/2e86f853bca54649c57205844fd33fcd4f0fc22c>.

src/lib/server/links/utils.ts
EthanThatOneKid: Attempted to resolve in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/607/commits/564efa820265f2df4fffa3ba7651ef6beb1d6e87>.

src/lib/server/links/utils.ts
EthanThatOneKid: Attempted to resolve in <https://github.com/EthanThatOneKid/acmcsuf.com/pull/607/commits/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb>.

src/lib/server/links/utils.ts
jaasonw: no jsdoc format?

src/lib/server/links/utils.ts
```diff
@@ -1,15 +1,76 @@
 import QRCode from 'qrcode';
-import { LINKS } from './data';
+import type { Link, ParsedLink } from './types';
 
-export function parseLinkId(link: string): keyof typeof LINKS | undefined {
-  const cleanedLinkName = link
+// parseLink parses the incoming link from a request.
```
jaasonw: had to read the source and tests to figure out what this do.
better comment probably something like "given a link, return the relative path name, query string, and hash", in which case how come not use built in url parsing https://developer.mozilla.org/en-US/docs/Web/API/URL/URL works on all major browsers, deno, and node

src/lib/server/links/types.ts
```diff
@@ -0,0 +1,10 @@
+export interface ParsedLink {
+  relativePathname: string;
+  query: string;
+  hash: string;
+}
+
+export interface Link<ID extends string> extends ParsedLink {
+  id: ID;
```
jaasonw: how come and ID that is a string and not just string

EthanThatOneKid: The type denoted `ID` is used to define `Link['id']`. We expect to use a simple JavaScript object to store the data for all of our official shortlinks. For example, let's say we have the following shortlinks represented in our shortlinks data.

<https://github.com/EthanThatOneKid/acmcsuf.com/blob/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb/src/lib/server/links/utils.test.ts#L4-L7>

We want the data we are working with to be as accurate as possible, so I used `ID extends string` to guarantee that only string-like types can be used to represent `ID`. In practice, the `ID` gets automatically inferred from the keys of the the official shortlinks collection, `LINKS`.

For instance, in the following line of code, `ID` is inferred to be represented by the keys of the test collection of shortlinks, `TEST_LINKS`, where `keyof TEST_LINKS` evaluates to `'example' | 'example/abc'` and is automatically used to represent `ID`.
<https://github.com/EthanThatOneKid/acmcsuf.com/blob/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb/src/lib/server/links/utils.test.ts#L212>

src/lib/server/links/utils.ts
EthanThatOneKid: I gave a shot at updating the content of the comment to be a bit more clear <https://github.com/EthanThatOneKid/acmcsuf.com/pull/607/commits/0403e9899046a9e746703e1d5731eab5d8007720>.

src/lib/server/links/utils.ts
EthanThatOneKid: Nice catch; I added JSDoc formatting.

src/lib/server/links/types.ts
EthanThatOneKid: > how come and ID that is a string and not just string

The type denoted `ID` is used to define `Link['id']`. We expect to use a simple JavaScript object to store the data for all of our official shortlinks. For example, let's say we have the following shortlinks represented in our shortlinks data.

<https://github.com/EthanThatOneKid/acmcsuf.com/blob/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb/src/lib/server/links/utils.test.ts#L4-L7>

We want the data we are working with to be as accurate as possible, so I used `ID extends string` to guarantee that only string-like types can be used to represent `ID`. In practice, the `ID` gets automatically inferred from the keys of the the official shortlinks collection, `LINKS`.

For instance, in the following line of code, `ID` is inferred to be represented by the keys of the test collection of shortlinks, `TEST_LINKS`, where `keyof TEST_LINKS` evaluates to `'example' | 'example/abc'` and is automatically used to represent `ID`.
<https://github.com/EthanThatOneKid/acmcsuf.com/blob/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb/src/lib/server/links/utils.test.ts#L212>

What this example does not go on to show is the usefulness of inference in TypeScript considering how the return type of `parseLinkId(data, TEST_LINKS)` will be of the appropriate type as inferred by the type of the keys of `TEST_LINKS`.

EthanThatOneKid: To clarify, Jason is asking why `Link` is defined like so:
<https://github.com/EthanThatOneKid/acmcsuf.com/blob/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb/src/lib/server/links/types.ts#L7-L10>

...When it could be defined like this:
```ts
export interface Link extends ParsedLink {
  id: string;
  destination: string;
}
```

src/lib/server/links/utils.ts
EthanThatOneKid: > how come not use built in url parsing https://developer.mozilla.org/en-US/docs/Web/API/URL/URL works on all major browsers, deno, and node

As far as I know, the `new URL()` constructor will throw an error if the string is not a valid URL. The "link" being passed into `parseLinkId` is expected to be the absolute pathname of the already parsed URL.
<https://github.com/EthanThatOneKid/acmcsuf.com/blob/697014c6ae2d4d6c4a095ab8b9f9d95e9e515adb/src/lib/server/links/shortener.ts#L7-L8>

This makes me realize that perhaps the current argument name perhaps is not sufficient. Perhaps `pathname` or `absolutePathname` instead of `url`. What do you think, @jaasonw?

src/lib/server/links/utils.test.ts
```diff
@@ -1,7 +1,215 @@
 import { test, expect } from 'vitest';
-import { parseLinkId } from './utils';
+import { parseLinkId, parseLink, findLinkId } from './utils';
 
-test('parses shortlinks', () => {
-  expect(parseLinkId('missing-link')).toBe(undefined);
-  expect(parseLinkId('GitHub')).toBe('github');
+const TEST_LINKS = {
+  example: 'https://example.com',
+  'example/abc': 'https://example.com/123/',
```
EthanThatOneKid: Quality-of-life feature idea: Support known shortlink IDs as a prefix for shortlink destinations.

Given the shortlink `https://acmcsuf.com/green`, we expect a computed destination of `https://example.com/green` where destination value `/example/green` of link ID `green` computes to (`/example` → `https://example.com`) + `/green`.

```suggestion
  example: 'https://example.com',
  'example/abc': 'https://example.com/123/',
  'green': '/example/green',
```

jaasonw (approved):

src/lib/server/links/utils.ts
jaasonw: yea sure

src/lib/server/links/utils.test.ts
EthanThatOneKid: Added support for internal redirect resolution in recent commits.

anguzz (approved):
🎃 spooktacular 👻

anguzz (approved):
letsgoooo 👻

src/lib/server/links/utils.ts
EthanThatOneKid: We ended up leveraging URL parsing in the latest version of this PR. Good suggestion!

PR #612, opened 2022-10-12
Set up Playwright screenshot testing

EthanThatOneKid:
### What we can do now

`npm t` has now been upgraded to include parallelized Playwright tests and screenshot diff testing. To specifically test the Playwright tests and not the Vitest tests, you may run `npm run test:browser`. Make some changes on screenshot-tested pages and give it a try!

```bash
npm t
```

### Change log
- Split up `/tests/test.ts` tests into each page of the `/src/routes/` directory.
- Added `projectsFromMatrix` helper to generate a matrix of screenshots to be taken per page.
- Added `vitest.config.ts` to exclude Playwright tests from the Vitest testing.
- Generated the first official set of screenshots in `/src/routes/screenshots/`.
- This PR goes on to add the screenshot testing boilerplate for each page endpoint of the website.
  - [x] Added screenshot tests for 404 error page.
  - [x] Added screenshot tests for `/1st` page.
  - [x] Added screenshot tests for `/about` page.
  - [x] Added screenshot tests for `/blog` page.
  - [x] Added screenshot tests for `/blog/:id` page.
  - [x] Added screenshot tests for `/events` page.
  - [x] Added screenshot tests for `/nodebuds` page.
  - [x] Added screenshot tests for `/privacy` page.
  - [x] Added screenshot tests for `/quiz` page.
  - [x] Added screenshot tests for `/teams` page.

Resolves #557.

PR #624, opened 2022-10-19
Removed screenshot testing from complex pages

EthanThatOneKid:
Screenshot testing was recently added in #612. It includes automatic screenshot testing for all `acmcsuf.com` pages despite not every page rendering deterministically. Screenshot testing should only be applied to pages that do not move much (which is expected behavior on a majority of our pages). For this reason, screenshot tests are prone to failure for the `/events`, `/blog`, and `/1st` pages.

Note: To support screenshot tests on pages with nondeterministic full-page screenshots, screenshots must instead be taken of individual page elements which are deterministic.


charliettaylor (approved):
LGTM!

Issue #546, opened 2022-09-13
Boardify: Board member picture frame generator


EthanThatOneKid:
### What's your idea?

@jaasonw suggested that we have a dedicated web application to facilitate the creation of board member pictures that are used on `/about`.

Context:
6:06 PM]jasonw: nah, like imagine like a web app
[6:06 PM]jasonw: u upload a headshot
[6:06 PM]jasonw: and then u move the picture around to fit
[6:06 PM]jasonw: and then when u hit export, it's in the correct size and format

Related: #348

k4rni:
Related: <https://github.com/EthanThatOneKid/acmcsuf.com/issues/348#issuecomment-1068831375>

PR #625, opened 2022-10-20
Fix blog image by fetching from Github first before relative path image

### LINKED ISSUE #619: Blog: Officer image fetched from incorrect path
### What happened?

The src attribute for this img element is not accurate. This is likely a simple oversight.

image

To clarify, we have a relative path when we want an absolute path.

anhduy1202:
# Description 
- Fixed Blog Author Image ( Ethan's image ) by fetching image from Github first 

# Result
image

### DIFF
--- a/src/lib/server/blog/posts.ts
+++ b/src/lib/server/blog/posts.ts
@@ -84,7 +84,8 @@ function cacheBlogPosts(output: any): BlogPost[] {
     const officer = getOfficerByGhUsername(author.login);
     const authorUrl: string = author.url;
     const displayname: string = officer?.fullName ?? author.login;
-    const picture: string = author.avatarUrl ?? officer?.picture;
+    const picture: string =
+      author.avatarUrl ?? `/assets/authors/${officer?.picture || 'placeholder.webp'}`;
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const labels = discussion.labels.nodes.map(({ name }: any) => name);
--- a/src/lib/server/blog/posts.ts
+++ b/src/lib/server/blog/posts.ts
@@ -84,7 +84,7 @@ function cacheBlogPosts(output: any): BlogPost[] {
     const officer = getOfficerByGhUsername(author.login);
     const authorUrl: string = author.url;
     const displayname: string = officer?.fullName ?? author.login;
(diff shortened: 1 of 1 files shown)


EthanThatOneKid (approved):
LGTM!!

src/lib/server/blog/posts.ts
```diff
@@ -84,7 +84,7 @@ function cacheBlogPosts(output: any): BlogPost[] {
     const officer = getOfficerByGhUsername(author.login);
     const authorUrl: string = author.url;
     const displayname: string = officer?.fullName ?? author.login;
-    const picture: string = officer?.picture ?? author.avatarUrl;
+    const picture: string = author.avatarUrl ?? officer?.picture;
```
EthanThatOneKid: The `officer?.picture` contains just the filename (e.g. `ethan-davidson.webp`).
```suggestion
    const picture: string = author.avatarUrl ?? `/assets/authors/${officer?.picture || 'placeholder.webp'}`;
```
Note: I am thinking that once we resolve #546, we can go back to using `officer?.picture` before `author.avatarUrl`.

anhduy1202: do u want me to commit that suggestion?

anhduy1202: > The `officer?.picture` contains just the filename (e.g. `ethan-davidson.webp`).
> 
> Note: I am thinking that once we resolve #546, we can go back to using `officer?.picture` before `author.avatarUrl`.

Actually, since our blog post is posted through Github and we also link our Github profile to the blog author, I think it's better to have the author picture same as Github pfp? thoughts?

EthanThatOneKid: > Actually, since our blog post is posted through Github and we also link our Github profile to the blog author, I think it's better to have the author picture same as Github pfp? thoughts?

We plan to store the GitHub usernames of most of our current board members in `officers.json`, so the we are technically able to then reference the board member's information. This includes information such as the board member's full name (used in a line of code above) and custom photo instead of their default GitHub profile picture. 

If the blog post author is a board member with a known GitHub username, we should be able to use the board member's headshot.

anhduy1202: make sense, pushing changes now

EthanThatOneKid (approved):

PR #637, opened 2022-10-24
Playwright testing: Remove screenshot testing

EthanThatOneKid:
### Background

The acmcsuf.com full page screenshot testing experiment has come to an end (for now). In its current state, the screenshot testing is like my hair: a bit long and cool, but always getting in the way! For this reason, I will delete the screenshot tests from the `main` branch but I will leave the individual Playwright page tests for future testing.

See: <https://github.com/EthanThatOneKid/acmcsuf.com/pull/612#issuecomment-1277076098>.

### Legacy

- Screenshot testing will be left as-is on the 404 page since I expect this page to change very rarely.
- Playwright tests will run on [all matrix combinations](https://github.com/EthanThatOneKid/acmcsuf.com/blob/ee1a0d9e7db95ee3178dec109addf06020d6a7f4/playwright.config.ts#L9-L20).


k4rni (approved):
This was an interesting test and it was great to see you play around with it. It's good to see how effective it was and by simply leaving it to test a singular page can be extremely useful!

PR #652, opened 2022-10-31
Install Git [Link to Download] - Getting_Started.md

### LINKED ISSUE #584: Provide download link for Git
We should probably provide a link to the actual download site for Git [in `GETTING_STARTED.md`].

_Originally posted by @karnikaavelumani in https://github.com/EthanThatOneKid/acmcsuf.com/pull/577#discussion_r977856574_

erwinmedina:
Added a new item [line 4] to the getting started list within the Getting_Started.md file. I added Item2: Installing Git [Link to Git/Downloads]. I also adjusted the numbers listed after that so it would read [1,2,3,4,5,6,7] instead of [1,2,2,3,4,5,6].
Link Issue: [Issue #584](https://github.com/EthanThatOneKid/acmcsuf.com/issues/584)

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -1,12 +1,13 @@
 ## Getting started
 
 1. Install [Node 16 LTS](https://nodejs.org/en/)
-2. Install and get familiar with [Git](https://docs.github.com/en/get-started/quickstart/hello-world)
+2. Install [Git](https://git-scm.com/downloads)
+3. Install and get familiar with [Git](https://docs.github.com/en/get-started/quickstart/hello-world)
    - If you are not comforable with the Git command line tool, you may use GitHub Desktop (<https://desktop.github.com/>). Take a look at <https://acmcsuf.com/1st/> if this is your first time contributing to an open source project.
-3. Create a fork of [acmcsuf.com](https://github.com/EthanThatOneKid/acmcsuf.com), similar to <https://github.com/jaasonw/acmcsuf.com>
-4. Create an account on [Vercel](https://vercel.com/) and create a new project linked to the fork you just created, similar to <https://acmcsuf-wayson.vercel.app/>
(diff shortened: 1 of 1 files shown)


k4rni (changes requested):
Thank you so much Erwin for taking a look at this! There are a couple things that you can edit to make this PR even better in fact. An interesting thing about markdown is its ability to create ordered lists without having a numerical ascending, descending or random order. 
image

What this means is that instead of labeling all of the steps as:
```md
1. 
2. 
3. 
```

We would rather change it so that every single number in the list will begin with just 1 since numbers are automatically rendered no matter the order used in the markdown:
```md
1. 
1. 
1. 
```

Leaving them all as '1' would make it so that future updates to this list are as simple as appending a new line to the list instead of having to edit all the numbers individually.

If you would like more elaboration, feel free to take a look at this: https://www.markdownguide.org/basic-syntax/#ordered-lists
[Basic Syntax | Markdown Guide](https://www.markdownguide.org/basic-syntax/)

GETTING_STARTED.md
```diff
@@ -1,12 +1,13 @@
 ## Getting started
 
 1. Install [Node 16 LTS](https://nodejs.org/en/)
-2. Install and get familiar with [Git](https://docs.github.com/en/get-started/quickstart/hello-world)
+2. Install [Git](https://git-scm.com/downloads)
+3. Install and get familiar with [Git](https://docs.github.com/en/get-started/quickstart/hello-world)
```
k4rni: I would also personally remove the "Install and" since it is already mentioned in the previous step to install Git

erwinmedina: That's a good call, i will make sure to remove that line, it was a little redundant. So sorry.

k4rni: Thank you so much!

k4rni (approved):
Thank you for all the help Erwin! Looks good to me:)

PR #657, opened 2022-10-31
Remove Vercel Audiences

EthanThatOneKid:
### Why?

1. Our previous Analytics were failing when in conjunction with the Audiences implementation.
2. There are no specific examples or instructions on implementing Vercel Audiences in a SvelteKit application.
3. One problem seems to be the mystery of the `/va/script.js` because I do not know where that is on the Internet, yet it seems to be required by `@vercel/analytics` despite finding zero documentation on the subject.

No analytics

### Future work

Perhaps I am missing something small, but I know that I must revert the Audiences changes in order to bring back our Vercel  Analytics and metrics. When new information is released by Vercel, it might be worth it to try reimplementing Vercel Audiences.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "3.0.0",
       "dependencies": {
         "@js-temporal/polyfill": "^0.4.1",
-        "@vercel/analytics": "^0.1.3",
         "html-to-text": "^8.2.1",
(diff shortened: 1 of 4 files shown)


k4rni:
lgtm!

k4rni (approved):

PR #566, opened 2022-09-15
Increased length of clickable space in navbar

k4rni:
Previously, in mobile mode, the size of the clickable area in the navbar was significantly small so it required you to press the names of the pages directly rather than the surrounding area itself.
Animation

Now it has been updated to encompass a larger area of clickable space. The size should be from one end of the screen to the other end.
Animation2

I am open to suggestions if whether we should have a title for the navbar, such as like "Menu", so that I can push down all the pages and make it easier to click if they are all situated near the bottom. I am also unsure of how to fix the words stacking on top of each other in "Node Buds". I am aware we have removed that page from the navbar, but for future pages it should not be stacked.

Fixes #279.

This PR additionally makes the navbar transparent, which addresses #580.

PR #667, opened 2022-11-02
Make sure the buttons render correctly on dark mode

### LINKED ISSUE #658: Button does not render correctly in darkmode
### What happened?

 Buttons do not render properly in dark mode without javascript. The text inside the button does not display properly
acm
acmbug2

vudiep411:
Buttons do not render correctly in dark mode because the text and the background color of the button are both white. Changing --button-bg: var(--acm-gray); ( Changing the background color of the button to gray) will do the trick.
Possibly because some browser does not load the javascript for "--acm-dark" in .global.css correctly. Fixes #658

### DIFF
--- a/static/global.css
+++ b/static/global.css
@@ -174,9 +174,9 @@ body:has(nav > input[type='checkbox']:checked) {
   --navbar-bg: rgba(16, 19, 21, 0.9);
   --footer-bg: var(--acm-gray);
   --card-bg: var(--acm-gray);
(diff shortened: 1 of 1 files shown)


k4rni (approved):
Thank you @vudiep411 for working on this issue! I actually encountered this issue in a PR I was working on (for reference: #566). 
image

In this PR, you can see where I made the mistake that you have corrected so I applaud you for your attention to detail. Thank you once again for your help!

Close #659.

PR #674, opened 2022-11-05
Fix non null assertion fail in Officer profile list

### LINKED ISSUE #641: Forbidden non-null assertion warnings everywhere!
### Ahhh!! 👻

image

### Valid non-null assertions

I believe non-null assertions should be used sparingly in TypeScript due to its ability to infer types like eating candy (too easy). However, in some cases, I believe it is appropriate to use it.

For example, the while-loop below ensures `next !== undefined`, but TypeScript is just not smart enough to know that. Thus, our valid usage of non-null assertions informs TypeScript that these values cannot be undefined.

<https://github.com/EthanThatOneKid/acmcsuf.com/blob/dfcf4d6d7205bada3f2018b59d9ffc67ccb51763/src/lib/server/links/resolve.ts#L19-L37>

### Suggestion

This could be remedied by updating the `.eslintrc.cjs` to include a custom rule `@typescript-eslint/no-non-null-assertion`. 

See: <https://typescript-eslint.io/rules/no-non-null-assertion/>.

anguzz:
Resolve 641

Co-Authored-By: diamondburned <diamond@arikawa-hi.me>

### DIFF
--- a/src/routes/(site)/about/officer-profile-list.svelte
+++ b/src/routes/(site)/about/officer-profile-list.svelte
@@ -24,9 +24,12 @@
    */
   function sortByTier(termCode: Term) {
     return (a: Officer, b: Officer) => {
-      const aTier = a.positions[termCode]!.tier;
-      const bTier = b.positions[termCode]!.tier;
-      return aTier - bTier;
+      const aPos = a.positions[termCode];
+      const bPos = b.positions[termCode];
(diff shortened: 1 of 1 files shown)


EthanThatOneKid (approved):
Looking over the shoulders of @Angus-1 and @diamondburned, can confirm that this passed tests 😎

PR #675, opened 2022-11-05
Fix properties of undefined regression in results screen

### LINKED ISSUE #673: Quiz: Regression in results screen 🙈
### What happened?

The results screen currently does not render. There is a JavaScript error to help us figure out what went wrong.

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'color')
    at Array.Dt (https://acmcsuf.com/_app/immutable/components/pages/(site)/quiz/_page.svelte-7a6075d4.js:5:4529)
    at Object.p (https://acmcsuf.com/_app/immutable/components/pages/(site)/quiz/_page.svelte-7a6075d4.js:7:8320)
    at Rt (index-81f88af5.js:4:1524)
    at mt (index-81f88af5.js:4:1210)
    at Ne (index-81f88af5.js:4:5182)
    at new jt (start-1420de02.js:1:9552)
    at Oe (start-1420de02.js:1:19413)
    at Object._hydrate (start-1420de02.js:1:30390)
    at async Mt (start-1420de02.js:1:31280)
```

vudiep411:
The keys in the JSON object are lowercase as shown below
quizconsole

So calling toLowerCase() function on: 
fix
(line 60 in quiz.svelte) will fix the undefined issue.
acmquiz

fixes #673

### DIFF
--- a/src/routes/(site)/quiz/quiz.svelte
+++ b/src/routes/(site)/quiz/quiz.svelte
@@ -57,7 +57,7 @@
   }
 
   $: talliedResponses = (responses ?? []).reduce((tallies, match) => {
-    match = match?.toLocaleLowerCase();
+    match = match?.toLowerCase();
     if (match && tallies[match]) tallies[match]++;
     else if (match) tallies[match] = 1;
     return tallies;
--- a/src/routes/(site)/quiz/quiz.svelte
+++ b/src/routes/(site)/quiz/quiz.svelte
@@ -57,6 +57,7 @@
   }
 
   $: talliedResponses = (responses ?? []).reduce((tallies, match) => {
+    match = match?.toLocaleLowerCase();
     if (match && tallies[match]) tallies[match]++;
     else if (match) tallies[match] = 1;
     return tallies;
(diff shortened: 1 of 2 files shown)


src/routes/(site)/quiz/quiz.svelte
```diff
@@ -57,6 +57,7 @@
   }
 
   $: talliedResponses = (responses ?? []).reduce((tallies, match) => {
+    match = match?.toLocaleLowerCase();
```
EthanThatOneKid: > toLocaleLowerCase() does not affect the value of the string itself. In most cases, this will produce the same result as [toLowerCase()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase), but for some locales, such as Turkish, whose case mappings do not follow the default case mappings in Unicode, there may be a different result.


In acmcsuf.com, we tend to use `toLowerCase` over `toLocaleLowerCase` since the method requires an argument to be of any difference to `toLowerCase`, which we practically never do.

```suggestion
    match = match?.toLowerCase();
```

vudiep411: changed toLowerCase()
acmlowercase

EthanThatOneKid: This works for me. Resolved.

EthanThatOneKid (approved):

PR #702, opened 2022-11-12
Added redirect button to 404 page and fixed flex gap

### LINKED ISSUE #688: Home Page Redirect on 404 page
### What's your idea?

Adding a redirect to the home page from the 404 for a smoother experience 🤠

j1yl:
# Description

## Changes
- Added <a> tag to 404 page
- Fixed flex spacing, removed margin bottom/tops from all elements, and added flex gap instead

### New
image
### Old
image

### DIFF
--- a/src/routes/+error.svelte
+++ b/src/routes/+error.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { page } from '$app/stores';
+  import Button from '$lib/components/button/button.svelte';
 </script>
 
 <svelte:head>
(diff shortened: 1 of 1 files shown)


src/routes/+error.svelte
```diff
@@ -9,13 +9,31 @@
 <section title={$page.error?.message}>
   <em>404</em>
   <h1>Frank can't find where you're going!</h1>
+  <a href="/">Return to Home</a>
```
EthanThatOneKid: For this scenario, we have a prebuilt common button component. Feel free to update our common button with your fine-tuned styles, just use the button imported like so:

```ts
  import Button from '$lib/components/button/button.svelte';
  ```

**Review by EthanThatOneKid — CHANGES_REQUESTED** [review]
Thanks for showing your interest by creating this PR! I left a comment below for you to address :)

**Comment by j1yl** [comment]
## Changes
- Removed custom button styles and changed to using Button component

## Issue
- When clicking the button, it opens the website in a new tab because of the following

```/src/lib/components/button/button.svelte```
```
<script lang="ts">
  export let link = '#';
  export let text = 'Submit';
</script>

<a class="brand-header size-md" role="button" href={link} target="_blank" rel="noopener norefferer">
  {text}
</a>
```

- the `target="_blank"` is causing the redirect to open in new tab

## Solution
- add prop to button component to specify whether target should be _blank or anything else

## Note
- this can be another beginner friendly issue on its own

src/routes/+error.svelte
j1yl: Added changes and found one issue, read comment below

jaasonw:
how come we dont just include the navbar in the 404 page?

jaasonw (changes requested):
also mousing over this button displays internal error 

image

j1yl:
Duplicate PR of already merged #718

PR #718, opened 2022-11-14
Added footer and navbar to layout on 404 page, add flex gap to 404 page

j1yl:
image
image

## Changes
- Added footer and navbar to `+layout.svelte`
- Added flex gap and removed margin top/bottom from `+error.svelte`

PR #721, opened 2022-11-15
Remove duplicate footer on all pages

### LINKED ISSUE #720: Footer appears twice on every page
### What happened?

The footer appears twice on every page. Happens on mobile and desktop.

DesktopExample

MobileExample

k4rni:
Upon approving PR #718, I mistakenly did not notice a duplicate footer was created as a result. This is caused by importing the footer and navbar to the wrong file `+layout.svelte` as opposed to `+error.svelte`. 

It is important to note that `+layout.svelte` is a blank layout. Stated in the comments, this file is used by the pages that don't need a layout and is also used as the base layout of all pages. 

As mentioned in the svelte docs (https://kit.svelte.dev/docs/advanced-routing#advanced-layouts), "Pages and layouts inside groups — as in any other directory — will inherit layouts above them". I presume that the `+layout.svelte` file above this one may have caused the duplicate footer, since it also imports the navbar and footer.
image

Fixes #720.

### DIFF
--- a/src/routes/+error.svelte
+++ b/src/routes/+error.svelte
@@ -1,7 +1,11 @@
 <script lang="ts">
   import { page } from '$app/stores';
   import Button from '$lib/components/button/button.svelte';
+  import Bar from '$lib/components/nav/bar.svelte';
+  import Footer from '$lib/components/footer/footer.svelte';
 </script>
+
+<Bar />
 
 <svelte:head>
   <title>ACM at CSUF / {$page.status || 404}</title>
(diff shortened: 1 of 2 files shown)


j1yl (approved):

j1yl:
So what we needed to do instead of adding a nav and footer to the `+layout.svelte` file was to add the nav and footer to the `+error.svelte` file. The previous PR caused a duplicate footer because there was already an import in the `+layout.svelte` in the `(site)` directory.

k4rni:
Yes @j1yl I believe that is what happened.

j1yl:
> 

Alright, seems like the issue is resolved. Thank you

k4rni:
Thank you for help as well!

PR #725, opened 2022-11-16
Remove split: true

### LINKED ISSUE #722: 404 page regression in Netlify migration
### What happened?

Our custom 404 page is currently not rendering as intended when visiting a page like <https://acmcsuf.com/this-page-does-not-exist>. 

500 Not found

I predict that additional configuration is required by Netlify to override their default error pages.

EthanThatOneKid:
> If this issue is considered serious u can always turn off split in the config file for a temporary solution. There r trade off in deploying SSR on netlify ig 🤷‍♂️

Suggested action sent via Discord by @vudiep411 😎

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -8,7 +8,7 @@ const config = {
   preprocess: preprocess(),
 
   kit: {
-    adapter: adapter({ split: true }),
+    adapter: adapter(),
   },
 };
 


jaasonw (approved):
look like it work
image

vudiep411 (approved):
Removing split will allow sveltekit to redirect to  +error.svelte page when a url is not matched.

jaasonw (approved):
?

PR #726, opened 2022-11-16
First contribution! Created some shapes.

DavidJSolano:
Edited and created three new shapes under acmcsuf.com\static\assets\1st\


.vscode/settings.json
```diff
@@ -5,5 +5,76 @@
...
+    "scoped_allocator": "cpp",
+    "shared_mutex": "cpp",
+    "sstream": "cpp",
+    "stdexcept": "cpp",
+    "streambuf": "cpp",
+    "thread": "cpp",
+    "typeindex": "cpp",
+    "typeinfo": "cpp",
+    "valarray": "cpp"
+  }
```
EthanThatOneKid: I am not sure how these changes got here lol. Go ahead and remove all your changes from this file since your PR does not require you to edit this file.

src/lib/public/confetti/new-shapes.ts
```diff
@@ -0,0 +1 @@
+export default 'url(/assets/1st/basic_shapes.svg)';
```
EthanThatOneKid: Nit: In acmcsuf.com, we tend to use [kabob case](https://en.wikipedia.org/wiki/Kebab_case) for storing our files.
```suggestion
export default 'url(/assets/1st/basic-shapes.svg)';
```

EthanThatOneKid (changes requested):
Thanks for the awesome shapes. I left you a couple of comments 😎

.vscode/settings.json
DavidJSolano: I'm just as clueless as you are with how this got there, just fixed tho thanks for pointing it out.

EthanThatOneKid (approved):
Thanks for making the requested changes! LGTM 😎

Issue #719, opened 2022-11-14
404 Redirect Button should not open a new tab


k4rni:
### What happened?

Since we added a new button on the 404 page to redirect the user to the home page, it instead opens up a new tab. This can be deemed as a suspicious pop up, therefore should be instead changed to redirecting on the current page.

![navbar2](https://user-images.githubusercontent.com/60043611/201770867-12bb6f44-e68c-4719-99b9-03bfd921dfd6.gif)

Refer to issue #718.

PR #730, opened 2022-11-20
Redirected 404 Button to stay on the original tab.

DavidJSolano:
First bug squashed. This PR fixes issue #719

### DIFF
--- a/src/routes/+error.svelte
+++ b/src/routes/+error.svelte
@@ -14,7 +14,7 @@
 <section title={$page.error?.message}>
   <em>404</em>
   <h1>Frank can't find where you're going!</h1>
-  <Button text={'Return to Home'} link={'/'} />
+  <Button text={'Return to Home'} link={'/'} redirect={'_self'} />
   <img src="/assets/png/lost-frank.png" alt="404 - Page Not Found" />
 </section>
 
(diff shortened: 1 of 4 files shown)


src/routes/+error.svelte
```diff
@@ -14,7 +14,7 @@
 <section title={$page.error?.message}>
   <em>404</em>
   <h1>Frank can't find where you're going!</h1>
-  <Button text={'Return to Home'} link={'/'} />
+  <Button text={'Return to Home'} link={'/'} redirect={'_self'} />
```
jaasonw: maybe change the behavior of this such that the prop is redirect=false to prevent opening in a new tab? imo better abstraction from the raw html

jaasonw: also i think default behavior should be to not redirect to new tab, there's only 3 files this button is imported in, can that be changed?

image

DavidJSolano: Got it, thanks for the suggestions. I just updated the PR so that the tweaks satisfy your requests. Let me know if anything else should be changed.

jaasonw (approved):
regarding the changes specifically i like it as is. personally i think it's kinda weird the button text is `<Button text="asdf">` and not `<Button>asdf</Button>` tho but thats out of the scope of this pr

PR #775, opened 2023-02-18
Remove `preventDefault` from anchor navigation

EthanThatOneKid:
I won't lie, I accidentally put `preventDefault` on the links in the navbar.


charliettaylor (approved):
I thought it was stop propagation but this works now, good job Ethan :-)
bob

k4rni (commented):
Thank you so much for catching this error and fixing it quickly!

EthanThatOneKid:
> Thank you so much for catching this error and fixing it quickly!

You caught it! It takes a village 🏘️

Issue #784, opened 2023-03-06
Google calendar to acmcsuf.com/events Gamedev, OSS, and ICPC colors are currently outdated


k4rni:
### What happened?

The ACM_TEAM="team" feature is currently outdated. What needs to be changed:

- [ ] ACM_TEAM=gamedev must be added
- [ ] ACM_TEAM=oss must be added
- [ ] ACM_TEAM=icpc must be added

**Note:**
Game Dev Team RGB values: (212, 17, 83)
Open Source Software Team RGB values: (17, 212, 177)
ICPC Team RBG values: (255, 121, 23)

<img width="698" alt="image" src="https://user-images.githubusercontent.com/60043611/223224636-0abd5702-e305-4576-8c14-589c05b221a4.png">

Reference #610 
Revisit #663

k4rni:
Resolved by #785, but with a slight typo followed up in #793.

PR #785, opened 2023-03-06
Add data for Game Dev, OSS, and ICPC team

k4rni:
### What changed?

- [x] Game Dev Team
    - Add data to `/src/lib/public/board/data/teams.json`.
    - Add `/static/assets/badges/gamedev-team.svg`.
    - Add additional case to `/src/lib/server/events/event.ts`.
- [x] OSS Team
- [x] ICPC Team

Incomplete resolution to #784


src/lib/public/board/data/teams.json
```diff
@@ -1,4 +1,16 @@
...
+    "title": "General",
+    "id": "general",
+    "picture": "/assets/badges/general.svg",
+    "color": "var(--acm-blue)"
+  },
+  {
+    "title": "AI",
+    "id": "ai",
+    "picture": "/assets/badges/ai-team.svg",
+    "color": "var(--acm-emerald)"
```
k4rni, line 12: @EthanThatOneKid Why do we use `"color": "var(--acm-emerald)"` instead of  `"color": "var(--acm-ai-rgb)"`?

EthanThatOneKid: > @/EthanThatOneKid Why do we use `"color": "var(--acm-emerald)"` instead of `"color": "var(--acm-ai-rgb)"`?

Our [`global.css`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/8bd8c6aa1aad39b43e1d90c13482a6b2605419e6/static/global.css) file includes several styles used globally throughout <https://acmcsuf.com/>. One section of this file is the `:root { /* ... */ }` block which instantiates the values of our global CSS variables.

In <https://acmcsuf.com/>, the variables with the prefix `--acm-` are all expected to be valid CSS colors. The variables with the suffix `-rgb` are all expected to be tuples of 3 integers (0-255) the first representing the amount of red, the second representing green, and the third representing blue. Note: This is not 100% consistent, so it is recommended to reference [`global.css`](https://github.com/EthanThatOneKid/acmcsuf.com/blob/8bd8c6aa1aad39b43e1d90c13482a6b2605419e6/static/global.css) each time we are working with a global CSS variable.

The variables with the prefix `--acm-` are intended to be used in CSS code that expects a color.

The variables with the suffix `-rgb` are intended to be used to set a custom opacity (presumably using the [`rgba`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb) syntax). For example:
<https://github.com/EthanThatOneKid/acmcsuf.com/blob/8bd8c6aa1aad39b43e1d90c13482a6b2605419e6/src/lib/components/toaster/toaster.svelte#L55>

This was a technique implemented quite some time ago, so I am open to any ways to improve our occurrences of `-rgb` CSS variables.

CC @karnikaavelumani

EthanThatOneKid (approved):
Exciting to see new branches added to `acmcsuf.com`!

Issue #804, opened 2023-03-11
Website Check: Update deprecated code


EthanThatOneKid:
> <img width="1380" alt="image" src="https://user-images.githubusercontent.com/8981287/224457581-346244c0-fde2-43ab-b9e5-087fdbb311a0.png">
> while you're at it can you update the github action : )

_Originally posted by @jaasonw in https://github.com/EthanThatOneKid/acmcsuf.com/issues/803#issuecomment-1464747874_

jaasonw:
<img width="1358" alt="image" src="https://user-images.githubusercontent.com/8981287/224459220-dd24d2bb-9168-43b3-aecb-ac81a0f7c67a.png">
on gitcored too thanks smile

jaasonw:
@karnikaavelumani https://github.com/EthanThatOneKid/acmcsuf.com/actions/runs/4389986907

k4rni:
Resolved with #805.

k4rni:
> <img alt="image" width="1358" src="https://user-images.githubusercontent.com/8981287/224459220-dd24d2bb-9168-43b3-aecb-ac81a0f7c67a.png">
> 
> on gitcored too thanks smile

Issue #803, opened 2023-03-11
package.json: Update test script


k4rni:
### What happened?

The test script currently combines the Vitest tests and the Playwright into one script. This is inconvenient since I find myself <kbd>Ctrl</kbd>+<kbd>c</kbd>'ing right after the Vitest tests finish running to avoid running the Playwright tests.

### Suggestion

Separate the `test` script into two new scripts. This requires editing the [Website Check workflow](https://github.com/EthanThatOneKid/acmcsuf.com/actions/workflows/check_website.yaml), though.

jaasonw:
<img width="1380" alt="image" src="https://user-images.githubusercontent.com/8981287/224457581-346244c0-fde2-43ab-b9e5-087fdbb311a0.png">

while you're at it can you update the github action : )

k4rni:
Resolved with #805.

PR #805, opened 2023-03-11
Website Check: Update deprecated code and test script

k4rni:
Resolves:
- #803 
- #804


EthanThatOneKid (approved):
LGTM; way to kill 2 birds with 1 stone!

PR #819, opened 2023-03-21
Resolve missing shortlink with given origin

### LINKED ISSUE #818: Shortlinks: Unspecified origin navigates to 404
### What happened?

It is impossible to set up a shortlink to navigate to an `acmcsuf.com` page without specifying the origin <https://acmcsuf.com>.

### Proposed behavior

Instead of `throw`'ing when a shortlink is "not found", the `resolve` function could reference a default origin string to navigate to the expected destination.

Related: https://github.com/EthanThatOneKid/acmcsuf.com/pull/816#issuecomment-1475794966

k4rni:
See title.

Fixes #818.


EthanThatOneKid (approved):
LGTM! 🚀

PR #841, opened 2023-04-26
add toggle expand positions

JustinStitt:
for /wave1 the `Expand All!` button isn't toggleable. 

Fixed.

expand-acm

**note: it requires JavaScript 🙄 (it did before but worth mentioning)**


src/routes/(site)/wave1/+page.svelte
```diff
@@ -8,9 +8,12 @@
   import PositionList from './position-list.svelte';
   import { POSITIONS } from './data';
 
+  let expanded = false;
   function expandAll() {
     const positions = document.querySelectorAll('.position');
-    positions.forEach((el) => el.setAttribute('open', 'true'));
+    if (expanded === true) positions.forEach((el) => el.removeAttribute('open'));
```
JustinStitt: semantics. 

probably should be one for loop with an `if` inside the block.

JustinStitt: not gonna fix tho xd123

src/routes/(site)/wave1/+page.svelte
```diff
@@ -8,9 +8,12 @@
...
+    else positions.forEach((el) => el.setAttribute('open', 'true'));
+    expanded = !expanded;
```
k4rni: Thank you so much for the review @JustinStitt! I appreciate you looking into this page.

I would like to offer a couple of suggestions:

1. You can simplify your code by using `if (expanded)` instead of `if (expanded === true)` since it is a little redundant despite it being easier to read. However, expanding the for each loop would make it much easier for beginners to read your code as the absence of curly braces and standard for loop structure may cause some confusion.
2. Another potential issue that could arise is when a user manually opens all the drop-downs and then tries to use the provided button to "close all." In such a scenario, the button must be pressed twice to achieve the desired outcome.

src/routes/(site)/wave1/+page.svelte
jjoeldaniel: resolved in #842

src/routes/(site)/wave1/+page.svelte
jjoeldaniel: 1 resolved in #842, [here](https://github.com/EthanThatOneKid/acmcsuf.com/pull/842/commits/2ae8197a84e81263754fb19d6fbcc088778b0255)

2 resolved [here](https://github.com/EthanThatOneKid/acmcsuf.com/pull/841/commits/a9c9090ee060e2c98ee764935f9bac7e9fd73531) and [here](https://github.com/EthanThatOneKid/acmcsuf.com/pull/841/commits/4d98f4d60ffd21f974ed4d90f09934fd57d027d7)

jjoeldaniel (approved):
after a thorough analysis and pep8 consult, i have come to the conclusion that this code does, in fact, lgtm

PR #859, opened 2023-05-22
Make Pomo Character's pixel art sharp

jiink:
Done by changing the filtering from auto to pixelated.

Before:

image

After:

image


k4rni (approved):
This looks a lot cleaner. I also learned this for the first time. Thank you for your PR!

EthanThatOneKid (approved):
Good catch, @jiink! Thank you!

PR #867, opened 2023-06-04
fix: confetti pointer event

### LINKED ISSUE #751: 1st: Top text unselectable
### What happened?

The confetti animation seems to cover the text at the top of the page such that the text is currently unselectable. This makes links to more information unreachable.

JustinStitt:
The text was not selectable and it now **is**.

Use the preview below **v** to check it out!

Fixes #751.


EthanThatOneKid (approved):
Nice catch! Thanks for the contribution :)

PR #868, opened 2023-06-04
fix: event summary, glow, margins

### LINKED ISSUE #832: Events: Event component minor regressions
### What happened?

- Event preview elements are always highlighted by default when they should only be highlighted when hovered over or opened.
- Spacing between the action buttons and the event description text is small.

image

JustinStitt:
Fixes #832 

Specifically, this patch fixes both the issues brought up [here](https://github.com/EthanThatOneKid/acmcsuf.com/issues/832#issue-1659242823)
> * Event preview elements are always highlighted by default when they should only be highlighted when hovered over or opened.
> * Spacing between the action buttons and the event description text is small.

Try it out with the preview available below **v**


EthanThatOneKid (approved):
Thanks for your effort on this, @JustinStitt!

PR #889, opened 2023-07-11
Update GitHub banner

k4rni:
The [acmcsuf.com/bug](https://acmcsuf.com/bug) page displays an outdated version of a banner previously made. This commit is to replace the previous banner with the newly updated version representing the creation of the acmOSS team!

OSS Banner


EthanThatOneKid (approved):
This looks SO sick!! capylove

PR #891, opened 2023-07-11
Modified maximum width of paragraphs in homepage #729

### LINKED ISSUE #729: Paragraphs on the index page are too wide
### What happened?

On the acmcsuf.com homepage, we have some paragraphs that are not entirely easy to read due to the extensive width of the paragraph elements.

tomasohCHOM:
I set the maximum width of paragraph elements to be 50ch so that the text doesn't become to wide as the screen size gets wider. I also switched the use of pixels to character units as we are dealing with text here. Fixed #729 

| BEFORE | AFTER   |
| ----- | ------- |
| Before (Wide Screen)  | After (Wide Screen)      |
| Before (Shorter Screen)    | After (Shorter Screen) |


EthanThatOneKid (approved):
Awesome; thanks for contributing! Nicely done PR 😎

k4rni (approved):
Thank you for the changes! Your PR is written very well. Keep up the amazing work:)

PR #906, opened 2023-09-10
Remove `legacyPicture`

### LINKED ISSUE #907: About: Visual Safari bug
### iPhone Safari bug

https://github.com/EthanThatOneKid/acmcsuf.com/assets/31261035/d24413e0-f754-4467-89c3-379fd0b9bc6f

Issue originally reported by @ashleyc417. 

### Research

A visual regression regarding legacy board member pictures on the `/about` page appeared ever since #899.

k4rni:
To fix the background colors of all images, I have decided it would be best to manually update all images to permanently remove the background color. In the future, we will have no shadow, but instead a border (specific to each team color), to fit consistently with the redesign theme.


EthanThatOneKid (approved):
The removal of the `legacy` property from `src/lib/components/board-member/board-member.svelte`, across all our board member data is satisfying. Congratulations on getting this to work!

image

PR #911, opened 2023-09-18
Fix unused tiers indices

### LINKED ISSUE #904: Unused tier indices
Taking a look at the codebase, I noticed the `TIERS` constant was not being employed as intended. Here is the culprit: we are currently sorting board members by tier ID when we should be sorting board members by the tier index associated with that tier ID. This mistake was not immediately clear due to the sort continuously exhibiting expected behavior, but not for the correct reason.

Edit (clarification): Tier IDs happen to be sorted in the same way as their respective indices. However, they might not always be that way. Hence the coincidence of showing expected behavior.

https://github.com/EthanThatOneKid/acmcsuf.com/blob/b5c26cb516a33fb5254f455f7fabbd999041ebfb/src/routes/(site)/about/officer-profile-list.svelte#L32

Suggestion:

```ts
return TIERS[aPos.tier].index - TIERS[bPos.tier].index;
```

k4rni:
Previously, the board was displayed in the order of their IDs, but this behavior was updated in this PR to reflect being displayed by the indices to allow fast updates to positions. 

Files were separated to provide clarity.

Fix #904.


EthanThatOneKid (approved):
I like the way you reorganized `src/lib/public/board/data/index.ts` ;)

EthanThatOneKid (changes requested):
Fix `.github/workflows/check_website.yaml` errors.

EthanThatOneKid (approved):
Thank you for the speedy follow-up.

PR #916, opened 2023-09-28
Edit algo slides link

k4rni:
The `acmcsuf.com` was accidentally appended to the alias as a mistake.


EthanThatOneKid (approved):
Interesting 🤔

LGTM 👍

PR #938, opened 2023-10-18
Fix website redesign hero.svelte

k4rni:
The hero component had a couple errors so this PR is to fix those. 

### Changes
- Removed `+layout.svelte` (removes duplicate footer)
- Adjusted margins 
- Removed brute forced code (clamp and max-width screen size)
- Fix multicolor text

### Screenshots
image
image
image


k4rni:
@EthanThatOneKid Can you properly check and see if the CSS is appropriately adjusting to all screen sizes? In comparison to the other components, I am noticing the behavior of the `hero.svelte` page is different from the others. 

@tomasohCHOM I noticed why the multicolor text was not working originally. The color spans across the entire screen in tablet mode so we need to add a `width` to fix it. However, I could not center it properly. I added you for review if you would like to take a crack at it!

tomasohCHOM:
@karnikaavelumani I will take a look at it right now and tell you how it goes!

EthanThatOneKid (approved):
Looks pretty much good to go! 

However this caught my eye. Why do we lose so many lines here?
9036E94D-B8EA-4910-9FC0-2E95B6CCF35E

PR #960, opened 2023-11-13
Fix all instances of gamedev logo

k4rni:
The previous gamedev logo had issues rendering in the svg format so it was manually fixed and replaced.

Previous:
image

Current:
image


EthanThatOneKid (approved):
Impressive fix! Nice work on this rather tedious task, @karnikaavelumani!

PR #988, opened 2023-12-02
Added 3% to margin of desc

### LINKED ISSUE #981: Add a margin on the description component on acmcsuf.com/6
### What happened?

To prevent the text from appearing directly on the edge of the screen, a margin or padding should be added to the text of the description component on the acmcsuf.com/6 page. This issue occurs on desktop/tablet mode and not necessarily on mobile mode so the changes applied should not affect mobile mode significantly.

image

EvanCPSC:
changed to 3%, here's multiple sizes for reference
image
image
image
image

Fixes #981

### DIFF
--- a/src/routes/(site)/6/desc.svelte
+++ b/src/routes/(site)/6/desc.svelte
@@ -47,6 +47,8 @@
       max-width: 1280px;
       margin: 0 auto;
     }
+  
+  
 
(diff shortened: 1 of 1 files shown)


EthanThatOneKid (approved):
Very nice 🥇

PR #1005, opened 2023-12-08
Don't Expose Yourself

JustinStitt:
## What
Change the blurb about the AI Team to not mention exposing yourself.

## Why
Just a bad reading experience.

## Now
It's better.


EthanThatOneKid (approved):
LOL good catch @JustinStitt!

PR #1009, opened 2023-12-28
add theme logic to error page

### LINKED ISSUE #971: Light/Dark mode toggle not working on 404 page
### What happened?

The light/dark mode toggle button on the 404 error page correctly toggles the moon/sun image on the button but does not actually change the theme of the page. I'm not that familiar with how that part is implemented, but I think it is because the error page is using the default layout that has nothing inside of it.

noahbei:
Fixes the issue mentioned in #971 
I don't know if the fix is the most optimal way to do it because I just put the code from the +layout.svelte in (site) into the outer +error.svelte


EthanThatOneKid (approved):
Can confirm (using the [preview](https://deploy-preview-1009--acmcsuf.netlify.app/404)) that toggling the color preference is now possible from the error page. Nice work, @noahbei!!! Thanks, man 😎

PR #1014, opened 2024-01-22
Eliminate lorem ipsum on landing page

### LINKED ISSUE #1013: Replace lorem text in landing page
### What's your idea?

Really neat comment from the capybaras, but I fear we have to temporarily replace these blocks of text for now (until we can get actual quotes).

I suggest we could replace this text with comments / descriptions from older versions of the website. For instance:

EthanThatOneKid:
Resolves #1013.


k4rni (approved):
LGTM!

PR #1015, opened 2024-01-23
Fixed overflow on teams component (landing page)

### LINKED ISSUE #1006: Overflow bug on homepage's teams component
### What happened?

Overflow error happens on the homepage between viewport widths of 792px and 479px. Ideally, the design team description should shift to the next row. One possible solution is to use the [flex-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap) property to make children automatically wrap around to a new row based on the size of its container, however this solution will need a minor refactor of the component since each row is rendered separately.

image

tomasohCHOM:
Aims to fix #1006 

https://github.com/EthanThatOneKid/acmcsuf.com/assets/112128328/8fbe46f3-6512-42a2-851c-c1caebe05bdc

As of now, this will fix the overflow caused by the teams component, but we might need to explore a way to contain width of the component and having two badges at medium screen sizes, while maintain the current look on desktop view.

### DIFF
--- a/src/routes/(site)/teams.svelte
+++ b/src/routes/(site)/teams.svelte
@@ -148,7 +148,7 @@
       grid-template-columns: 1fr 1fr;
       align-items: baseline;
       justify-content: center;
-      width: 70%;
+      width: 70ch;
       gap: 2em;
     }
   }
--- a/src/routes/(site)/teams.svelte
+++ b/src/routes/(site)/teams.svelte
@@ -121,10 +121,21 @@
     }
 
     section .hero-inner-container {
+      grid-template-columns: 1fr;
       gap: 2em;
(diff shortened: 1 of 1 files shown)


src/routes/(site)/teams.svelte
EthanThatOneKid, line 1: Consider using the `ch` unit instead of the `%` unit for the container’s width.

EthanThatOneKid (changes requested):

src/routes/(site)/teams.svelte
tomasohCHOM: Done!

EthanThatOneKid (approved):
Nice; LGTM! 😎

PR #1018, opened 2024-02-15
Fix toggle button on error page

### LINKED ISSUE #971: Light/Dark mode toggle not working on 404 page

EthanThatOneKid:
Fixes #971.


EthanThatOneKid:
Duplicate of #1009.

PR #1022, opened 2024-02-21
reflect new size-limit in contributing guide

amyipdev:
The contributing guide currently lists the size-limit as 50 MB. This is incorrect, as the size-limit is (and has been for some time) 256 MB. This patch updates the contributing guide to reflect that difference.

Signed-off-by: Amy Parker <amy@amyip.net>


EthanThatOneKid (approved):
Good catch! Thanks!

PR #1029, opened 2024-03-13
fix game dev anchor styling on about page

caretcakejake:
Sorry I didn't make an issue first y'all  😞 
- Fixes the issue where game dev anchor tags show up as dev tags on the about page

### Before
image

### After
image


k4rni:
Thank you for your PR jacob! We actually fixed this in another PR. 

Thank you for the bug checks:)

PR #1116, opened 2024-09-06
fix: dark mode 

### LINKED ISSUE #1114: Icons aren't visible on dark mode when events are created
### What happened?

Look at photo for reference:
image

tomasohCHOM:
Fixes #1114 

image

These icons were affected because of a change to how we handled dark mode. In order to take advantage of SSR and prevent theme flickering on page load, we used html data attributes and cookies to persist preferences. That means we no longer needed to have the `.dark` and the `.light` class names, which I forgot to check for other places where the codebase used those.

Switched from using `body.light` and `body.dark` to `html[data-theme='light'` and `html[data-theme='dark'` instead.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1326,11 +1326,11 @@
       }
     },
     "node_modules/axobject-query": {
-      "version": "3.2.1",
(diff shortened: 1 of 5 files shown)


26samaahmed (approved):
LGTM!

PR #1143, opened 2024-10-29
Fixed blog syntax highlighting not working

### LINKED ISSUE #1092: Blog: Syntax highlighting is broken
### What happened?

Blog post syntax highlighting is not working! 😢 
image

tmbkoren:
Wrapped every css selector in syntax-highlight.css in :global(), which solves the issue without breaking anything else. It is probably a bandaid solution, but I don't see any other way without rewriting the whole blog system.


tomasohCHOM (approved):
LGTM

PR #1146, opened 2024-11-02
Add Spin Effect to Hero Section

### LINKED ISSUE #1124: Add Circular Spin Effect to Homepage Team Icons
### What's your idea?

it would look REALLY cool

**S P I N**
image

dianellasy:
I worked on creating a spin effect on the home page with hero and the logos of the ACM teams. 

https://github.com/user-attachments/assets/cb925660-b4b9-4473-8b57-1b0fd1155e6b


26samaahmed:
@dianellasy It looks awesome! I think it would be better if we have the capybara not spin with the logos. The current issue that's preventing us from doing so is that the capybara and logos are linked in 1 image. I will separate them to 2 different images that are layered on top of each other and then we can see if we're able to have only the logos spin! This resolves issue #1124

MarkRyanGarcia:
should be working in desktop and mobile view :D

https://github.com/user-attachments/assets/02f4ae74-d651-4878-824e-bc4933575572

26samaahmed (approved):
LGTM! Thank you @MarkRyanGarcia and @elenav24 for finishing the issue!

PR #1154, opened 2025-01-27
Add Node Buds to Landing Page + Change Formation

26samaahmed:
I changed the formation of teams to display 4 teams per row instead of 3 and added node buds team!

Screenshot 2025-01-26 160724


tomasohCHOM:
4 columns is too much and it will overflow when screen size is smaller than 1024px. I can quickly refactor this to keep the new layout and make it look better at smaller screen sizes.

tomasohCHOM (approved):
Keep it at 4 columns at large screen sizes, decrease for smaller screen sizes

tomasohCHOM (approved):
LGTM

PR #1160, opened 2025-02-20
Allow calendar entries with empty event descriptions to be displayed

### LINKED ISSUE #1158: Handle empty Google Calendar descriptions
### Handle empty Google Calendar descriptions

Currently, if a Google Calendar event lacks a description, it will not be displayed. While a majority of events contain a description, supporting events with blank descriptions would still be a nice addition. 

If someone has an opposing view, also willing to hear that out.

Having something extra that adds team descriptions to a teams event (ex. checking if "game dev" is in the title and then adding the teams info) could be neat, but not sure how feasible that is.

https://github.com/EthanThatOneKid/acmcsuf.com/blob/e906b6105bfe683dcc728a059ed2b389942cd5c5/src/lib/server/events/event.ts#L5-L16

PillowGit:
Fixes #1158 

Instead of completely denying all calendar events from the acmcsuf public calendar that don't have a description, we display them on the site with a little message so that the event time and title are still visible to those interested.

```js
  const { description, variables } = parseDescription(
    event.description || 'Event description is missing :('
  );
  ```


NateGries1:
It may look unprofessional to have that message show on our event. Imo it shows that we forgot to put one.

PillowGit:
> It may look unprofessional to have that message show on our event. Imo it shows that we forgot to put one.

Are you suggesting a different message? Or not having empty descriptions at all? If you don't think we should have events with empty descriptions at all then please mention it on the #1158 discussion instead since this pr is for that issue

dianellasy (approved):
LGTM!

PR #1162, opened 2025-02-21
Fix gamedev syntax highlighting on marketing team

### LINKED ISSUE #1153: Incorrect Team Color on Teams Page
### What happened?

If an officer is assigned to both the 'Game Dev' and another team (e.g., 'Marketing'), the color for 'Game Dev' is not correctly applied and instead, the 'Dev' color is mistakenly used. This happens because the team titles get split, and the 'Game' and 'Dev' parts are treated separately. This results in the officer being displayed with the incorrect color for 'Dev', even when they should be shown with the color for 'Game Dev'. Look at pictures below for reference.

Image
Image

k4rni:
The previous if statement would only work if the current team was 'gamedev' and the current iteration check was 'dev', which doesn't work when on marketing or other teams.

This fix tweaks the if statement to specify the phrase "game" to ensure that the highlighting would be on the appropriate team.
image

Fix #1153.

### DIFF
--- a/src/routes/(site)/teams/officer-profile.svelte
+++ b/src/routes/(site)/teams/officer-profile.svelte
@@ -28,7 +28,7 @@
   }
 
   $: titleHTML = Object.values(TEAMS).reduce((s, t) => {
-    if (t.id === 'dev' && team?.id === 'gamedev') {
+    if (t.id === 'dev' && officerPosition.includes('Game')) {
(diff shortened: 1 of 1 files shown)


26samaahmed (approved):
Looks good! Thank you for working on this @karnikaavelumani @NateGries1!!

PR #1168, opened 2025-03-12
Add Node Buds Logo to Hero Section

26samaahmed:
I replaced the ICPC logo with the Node Buds Logo. In the future, we could have both teams but for now, we want to showcase the node buds board if people click on their logo. 
Screenshot 2025-03-11 183055


tomasohCHOM (approved):
LGTM!

PR #1184, opened 2025-06-29
remove update-genuary workflow

tomasohCHOM:
Removes `update-genuary.yaml` which is set to run every day. Consistent workflow failure is primary reason for removal


dianellasy:
Looks good to me!

dianellasy (approved):
Looks good to me!

PR #1188, opened 2025-07-04
Add Fall 2025 Board Members

dianellasy:
I added a new term on acmcsuf.com/teams for Fall 2025 and added the board members for each team. Also, when first opening /teams, it defaults to Fall 2025.
image
image


EvanCPSC (approved):
LGTM!

-🥪

PR #1166, opened 2025-03-05
Add Deno Deploy support

EthanThatOneKid:
### Changes

- Remove lc-dailies frontend.
- Remove workflows Update Deps, Update Genuary.
- Update outdated package.json dependency versions.
- Migrate to latest eslint and prettier configuration best practices.
- Change SvelteKit adapter to `adapter-node` for Deno Deploy.

### Methodology

This pull request implements an upgrade to the project's SvelteKit boilerplate by recreating the project from scratch using the latest `npx sv create` (or `deno -A npm:sv create`). This approach, while manual, ensures alignment with current SvelteKit best practices and eliminates potential technical debt accumulated from incremental updates. By directly integrating the new boilerplate, we address implicit changes in recommended project structure and configurations that may not be fully documented in migration guides. This method provides a clean slate for adopting the most recent project conventions, resulting in a more maintainable and performant application.

PR #1195, opened 2025-09-05
chore: update to svelte5

TheJolman:
- I originally did this in an attempt to get Deno Deploy early access to work (it allegedly has better npm support now) since its free tier supports github orgs.
	- Full disclaimer I just told a cli agent to iterate on this until the errors went away, so I'm not sure how sound this is but it *seems* to work.
- Should probably update to svelte 5 at some point, regardless of where the site is deployed.

### Also
- Added `"engines"` section to `package.json` that sets the required versions of nodejs and npm. These are set to `>=22.0.0` and `>=10.5.1` respectively. NodeJS 22 is the latest LTS version and these versions shouldn't be too restrictive. See https://nodejs.org/en/about/previous-releases for more info on versions.
- Created a `.nvmrc` file that references NodeJS v22.19.0. This is the latest patch for NodeJS 22 and will be installed by anyone using `nvm` and also used by the github action.
- Updated `website_check` github action:
	- bumped versions of all workflows
	- workflow references `.nvmrc` to get correct nodejs version
		- set version to `22.19.0` (latest LTS release)
	- Fixed usage of matrix.os
- Modernize ESlint config: switch to `eslint.config.js` from `eslint.cjs`


EthanThatOneKid:
Prior art: #1166

TheJolman:
Even if we stay on Netlify, updating the site to use Svelte 5 should probably be done sooner rather than later.

TheJolman:
dang this ended up being a rabbit hole but all seems to be good now

TheJolman:
getting a build failure on the windows runner only during 'Test Dependency Size', not sure how to proceed

TheJolman:
ok workflow issues all seem to be resolved

dianellasy (approved):
LGTM!

package-lock.json
EthanThatOneKid, line 1: Run `npm i` one more time.

EthanThatOneKid (approved):
@TheJolman upgrades the project to Svelte 5 and also updates the surrounding CI infrastructure to the latest versions. LGTM :godmode: 

Suggestion: Make sure the `package-lock.json` is up-to-date by running `npm i`.

src/lib/components/recursive-ul/recursive-ul.svelte
```diff
@@ -5,7 +5,7 @@
 </script>
 
 <ul class="recursive-ul">
```
EthanThatOneKid, line 7: Encountered warning:
```
node_invalid_placement_ssr: `<ul>` (src/lib/components/recursive-ul/recursive-ul.svelte:7:0) cannot be a child of `<p>` (src/routes/(site)/wave2/position.svelte:32:4)
```

TheJolman: think I just resolved this, can you try to reproduce this again?

package-lock.json
TheJolman: :+1: also I'm thinking of just removing the now outdated `bun.lockb`, do you know anyone that's actively using bun to work on this?

EthanThatOneKid: > do you know anyone that's actively using bun to work on this?

No lol

PR #1198, opened 2025-09-10
Update Discord Invitation

26samaahmed:
This is a small nitpick but I updated the discord server invitation because the current one says 'Deleted user' is inviting you. The person who had originally made the invitation deleted their profile so this will look a bit cleaner.


dianellasy (approved):
LGTM!

PR #1199, opened 2025-09-10
Fix Join Button Redirection

26samaahmed:
Join button on /events for every event should now redirect to the acm discord server. The assumption for why it didn't work previously is because something was being overridden when using the anchor tag so it's now replaced with a button tag that redirects to the server.


dianellasy (approved):
LGTM queen! Thank you for all of your hard work for the join buttons 🌟

PR #1202, opened 2025-09-19
Added a New Tier Named ICPC Lead

dianellasy:
I added a new tier named ICPC Lead and assigned Noah Scott to it.
image


EvanCPSC (approved):
LGTM! W change

PR #1207, opened 2025-09-21
Update ID for ICPC Lead Role

### LINKED ISSUE #1205: ICPC Lead is Displaying Under the Node Buds Team
### What happened?

After adding the ICPC Lead, Noah is now displaying under the Node Buds team.
Image

26samaahmed:
The reason the ICPC Lead was showing along the nodebuds team is because both roles had the same id, 34. To fix this, I added a unique id for the role, which is 27! 

Screenshot 2025-09-20 at 9 24 02 PM


dianellasy (approved):
Thank you so much!

PR #1215, opened 2025-10-25
Feat/workshop page

GaballaGit:
PR for Workshops Pages

The workshops page is meant to make it convenient to find workshops that have happened in the past. 

For now the workshops will be added manually. I convert links.json to a workshops table with a script I have and plan to do so once a week until we have a way to dynamically do so. Here is a link to the script:
https://github.com/GaballaGit/wstable-gen/tree/main 

Here is a preview of Workshop pages

https://github.com/user-attachments/assets/c90e6c4d-3685-43d2-88d2-b2e0c2b5e35c


TheJolman:
TBH I'm not sure, @dianellasy might know the answer though. Also in case you didn't notice `npm run check` is failing in CI

GaballaGit:
Alright, thanks for the heads up. I saw the failed checks and I think I will work on them tomorrow.

TheJolman:
Also looks good so far - but IMO using dropdowns would be a little sleeker. Struggling to find a good example rn but instead of a table for each branch it would be a list of wide buttons like `Fall 2024 (6) 🔽` and clicking it would reveal the titles of each workshop, hyperlinked with the link to the google doc. Just my opinion, thoughts?

GaballaGit:
It's an interesting idea. Right now I went off of what Ethan sent me (https://acmcsufoss.github.io/docs/workshops.html).

I am a little hesitant on switching to a drop down menu because if we decide in the future to add things like descriptions to a workshop it could get messy in drop down format.

GaballaGit:
With this recent commit I think I want to open the PR for a review. Please take a look whenever you are available @dianellasy @EvanCPSC.

EvanCPSC (approved):
LGTM!

PR #1226, opened 2025-12-01
Feat/1193

### LINKED ISSUE #1193: Redesign the /Hackathons Page
### What's your idea?

Currently, the design for /hackathons is not appealing because there is a lot of content in the middle that can be displayed better. For example, displaying two hackathons right next to each other and changing the emojis to match the hackathon's theme.

https://github.com/user-attachments/assets/cb1617a7-4977-4ca8-80c7-76fad35a08ce

bebopkenny:
I made a title to describe what the page is about with pills so users can select a year of a specific hackathon they want to go to. Hackathon images pop up when hovering over them, similar to the director cards. While scrolling through the sections, the background color changes, similar to how it is on the ACM Teams page. I changed the order of which year comes first. I sorted it so the most recent year appears first, then goes down the list. I made a back to the top button when users are on the bottom of the webpage, they can simply click the button to get back to the top of the page. Each section does have a unique emoji that represents the theme compared to before, when they all shared the same emoji. 

https://github.com/user-attachments/assets/977aa293-e625-41fa-b4e1-b589a21dcd8e


dianellasy:
Hi @bebopkenny!

I like the redesigned ```/hackathons``` page! It is very clean. However, I have some changes to make it better!

1. When you click on the years at the top of the page, the scroll effect should scroll to the year's hackathon (which it already does), but it should show the title "FullyHacks *Year*". For example, if you click "2024", it should scroll to "FullyHacks 2024" and show the title.

https://github.com/user-attachments/assets/5ad3bf63-f9d0-48b5-93e5-c324bee09067


- When you hover over the FullyHacks picture, remove the "click" icon. The "click" icon should only appear if the item is clickable, which, in this case, it is not. 

https://github.com/user-attachments/assets/35fe79ea-6f03-47ec-8a21-4fb14a4718c6

Overall, this is really good and way better than the previous ```/hackathons``` page!

bebopkenny:
I made the following changes as requested: "When you click on the years at the top of the page, the scroll effect should scroll to the year's hackathon (which it already does), but it should show the title 'FullyHacks Year'. For example, if you click '2024', it should scroll to 'FullyHacks 2024' and show the title. When you hover over the FullyHacks picture, remove the 'click' icon. The 'click' icon should only appear if the item is clickable, which, in this case, it is not." Also flipped the background color scheme so the first section starts with the lighter blue color then switches to the darker blue. Added the spacing component from the Teams webpage to keep the last section separate from the footer.

https://github.com/user-attachments/assets/c9a071fa-c40b-4a58-abec-61361718d7a7

dianellasy:
Hi @bebopkenny!

I ran the branch locally, and for some reason, it still does not show the hackathon title on my computer or @26samaahmed's when I click the buttons of the years at the top of the page.

Also, I think it is unnecessary that when you click on the title, such as "FullyHacks *Year*", it scrolls to the image. The image is already shown underneath, so I think that should be removed, as well as the light blue hover effect for the title.

Additionally, I believe that the hackathon description underneath each image should span the entire image.

dianellasy:
Additionally, I added a scroll-to-top component and placed it on the home, teams, and blog pages. 

I also moved branding to Board Use on /sitemap.

dianellasy (approved):
LGTM! Thank you so much @bebopkenny for working on this issue!

bebopkenny:
Pleasure is all mine. Thank you for your guidance. I’m always happy to help improve the ACM website!

PR #1231, opened 2026-01-05
remove duplicate files

TheJolman:
these duplicated files have no diff between their originals and seemingly do nothing. They appeared two months ago


EvanCPSC (approved):
lgtm ty for catching this

PR #1238, opened 2026-01-12
simplify cloudflare preview

TheJolman:
Instead of having two workers in the dashboard (acmcsuf-com-production and acmcsuf-com-preview), this simplifies things by removing the latter worker. `npm run preview` now uses a different command, `wrangler versions upload -e production` that deploys using the same configuration as deployments to the acmcsuf.com domain, but on a workers.dev domain with optional tag (see comment in wrangler.toml). View all deployments in the Deployments tab of the worker.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "start": "npm run dev",
     "dev": "wrangler dev",
     "build": "vite build",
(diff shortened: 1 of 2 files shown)


EvanCPSC (approved):
lgtm!

PR #1239, opened 2026-01-13
remove depreciated setting from `.npmrc`

TheJolman:
`resolution-mode=highest` is a depreciated setting. This removes it and also adds the `fund=false` setting (hides the "packages looking for funding" message)


dianellasy (approved):
LGTM!

PR #1241, opened 2026-01-14
fix /colors page by importing assets instead of fetching them

tomasohCHOM:
## What Happened

The `/colors` page has been broken ever since we migrated to Cloudflare Workers, failing when fetching `/global.json` with a 522 status. It could be the result of a infinity call cycle resulting in a timeout, but Ii's still not very clear what the root cause is with the lack of logs. This fix deviates from the previous approach, essentially not addressing the root cause but doing something different.

## Description of Fix

Instead of making an API request to fetch this resource, these changes simply move the location of `global.json` to `src/lib/assets/global.json` and then importing that as type `Color[]`. We get rid of the overhead from the API call and we also get some extra gains since Vite does some optimizations under the hood when [importing assets](https://svelte.dev/docs/kit/images#Vite's-built-in-handling). We could even get rid of `colors/+page.ts` and simply import the JSON file into the Svelte component, but will let the reviewer of the PR decide if I should or not.


dianellasy (approved):
LGTM!

PR #1246, opened 2026-01-19
use cloudflare vite plugin to re-enable `vite dev`

TheJolman:
When we migrated to workers, I changed the dev server command from `vite dev` to `wrangler dev` due to some inconsistencies between the vite dev server and production environment. However, we lose hot reloading with this change which is quite annoying.

The solution *appears* to be installing and using the official cloudflare vite plugin, which should give us the best of both worlds, and while using `vite dev` after installing works, `vite build` starts failing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "temporal-polyfill": "^0.3.0"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


TheJolman:
I don't think this is possible as of now. Closing in favor of #1249

PR #1278, opened 2026-05-04
improve cloudflare docs

TheJolman:
This original doc was like 80% AI generated and I don't like how it read. Took some time to pare it down to the most important parts and hopefully make it more readable and useful.


dianellasy (approved):
LGTM! Thank you for improving the Cloudflare docs!

PR #1280, opened 2026-05-14
modernize prettier config

TheJolman:
Moved config to `prettier.config.ts` to be more inline with other files.
Removed `.prettierignore` as prettier automatically respects `.gitignore`.

Lowk just did this bc the discrepancy between this and the other config files was bugging me.


GaballaGit (approved):
nice nice

PR #1283, opened 2026-07-08
Teams script

GaballaGit:
Introduces a script to help make the process of editing the board's officer data a lot easier. The script can be ran by cd'ing into the directory and using the following command: `node main.js`

The script creates a copy json of the current officers.json. Then it writes changes to it, and finally commits it to the original json.

It comes with several commands, they are explained in the comment block of main but I will also paste them here for convenience:

 > 1 -> Find officer. Type an officers name and their info will be displayed.
  [NOTE] The data that is displayed is the edited one, not the one currently present in production.
 
  > 2 -> Edit officer. Provides a list of prompts to go through to edit/add an officer.
 
  > 3 -> View current diff. The script writes to a copy of the board json before commiting changes to avoid mistakes.
  You can compare your current changes to what is on the board json with this.
 
  > 4 -> Commit changes. When you are confident and done with your changes to board, you can then write it to the original board json.
 
  > 5 -> Exit. Exiting will not save your progress (As well as hitting Ctrl + C in the middle of using the script).
 

Now you might notice a massive diff with the current officers.json file. It is because all the json has been expanded. That is because the way `JSON.stringify(content, null, 2)` outputs json like so. Everything operates as normal though.


GaballaGit:
Fails because of outdated test

dianellasy (approved):
LGTM!

PR #1294, opened 2026-08-19
fix @username page crash

bebopkenny:
Clicking @[username] on /sitemap was breaking the other links on page. The load function was parsing the response w/o checking if the request worked, so  when the endpoint returned a 500 page crash. Now it checks first and throws a proper 404 error page. 

The 500 still shows in the console. I think the github pat is expired, I tested it with my own pat and the page loaded fine.


src/lib/server/gql/github.ts
```diff
@@ -3,10 +3,26 @@ import { GH_ACCESS_TOKEN } from '$lib/server/env';
...
+
+  if (!r.ok) {
+    throw new Error(`GitHub query failed: ${r.status} ${r.statusText}`);
+  }
+
+  // A failed query responds with an errors array and no data.
+  const body = await r.json();
+  if (!body.data) {
+    const reason = body.errors?.map((e: { message: string }) => e.message).join('; ');
+    throw new Error(`GitHub query failed: ${reason ?? body.message ?? 'no data returned'}`);
```
GaballaGit: Already read the body message in response variable. Consider just doing
```suggestion
    throw new Error(`GitHub query failed: ${reason}`);
```
looks a lot cleaner and readable imo.

src/lib/server/gql/github.ts
```diff
@@ -3,10 +3,26 @@ import { GH_ACCESS_TOKEN } from '$lib/server/env';
...
   });
-  return (await r.json())['data'] as T;
+
+  if (!r.ok) {
+    throw new Error(`GitHub query failed: ${r.status} ${r.statusText}`);
+  }
+
+  // A failed query responds with an errors array and no data.
+  const body = await r.json();
+  if (!body.data) {
```
GaballaGit: A failed GraphQL has an error field you can look rather than checking if body data doesn't exist. Then you can just skip the ternary operator for reason below.
```suggestion
  if (body.error && body.error.length > 0) {
```
(make sure error exists with a len more than 0)

src/lib/server/gql/github.ts
```diff
@@ -3,10 +3,26 @@ import { GH_ACCESS_TOKEN } from '$lib/server/env';
...
     headers: { Authorization: `token ${GH_ACCESS_TOKEN}`, 'Content-Type': 'application/json' },
     body: JSON.stringify({ query }),
   });
-  return (await r.json())['data'] as T;
+
+  if (!r.ok) {
+    throw new Error(`GitHub query failed: ${r.status} ${r.statusText}`);
+  }
+
+  // A failed query responds with an errors array and no data.
```
GaballaGit: Not needed in this case, also isn't true all the time, consider just ending with 'an errors array we handle' or something similar.
```suggestion
  // A failed query responds with an errors array we handle.
```

GaballaGit (changes requested):
Looks good so far, thank you Kenny! I do have some suggestions I want you to take a look at first before merging though, related to error handling the failed graphql query.

src/lib/server/gql/github.ts
```diff
@@ -3,10 +3,26 @@ import { GH_ACCESS_TOKEN } from '$lib/server/env';
...
+  const body = await r.json();
```
GaballaGit, line 20: error handle if body doesn't exist so we don't access null at runtime.
```suggestion
  const body = await r.json();
  
  if (!body) {
      throw new Error(`GitHub query failed: no body returned`);
  }
```

GaballaGit (changes requested):

bebopkenny:
Thanks for the suggestions! I applied them right now. I dropped the body.message fallback and moved the comment down to where the errors check actually is.

src/lib/server/gql/github.ts
```diff
@@ -17,11 +17,16 @@ export async function doQuery<T>(query: string): Promise<T> {
...
   const body = await r.json();
-  if (!body.data) {
+
+  if (!body) {
+    throw new Error(`GitHub query failed: no body returned`);
+  }
+
+  // A failed query responds with an errors array we handle.
+  if (body.errors && body.errors.length > 0) {
     const reason = body.errors?.map((e: { message: string }) => e.message).join('; ');
```
GaballaGit: One more thing! Since we checked if body.errors must exist in the conditional, the ternary operator isn't needed.
```suggestion
    const reason = body.errors.map((e: { message: string }) => e.message).join('; ');

```

GaballaGit (changes requested):

GaballaGit (approved):
amazinggg!

PR #1296, opened 2026-08-27
replace shortlink resolver with s.acmcsuf.com

TheJolman:
replace shortlink resolver to s.acmcsuf.com. This uses the new shorter2 discord bot (live). Use the preview link below to test, all old links should have been migrated over. Only difference I'm seeing so far is that the links on `/shorter` are listed in a different order.


GaballaGit (approved):
gorgeous PR. Approved

